PR #384, opened 2022-04-05
Expose metrics from plugins

shounakmk219:


cyrilou242:
hey @shounakmk219 should thie be closed?

PR #412, opened 2022-04-19
[server] Endpoint to handle orphan tasks. Multi worker setup.

shounakmk219:
When a worker is terminated while it's running a task the task is stuck with the `RUNNING` status until a worker with same id boots up and marks it as `FAILED`. 
As we are assigning random ids when a worker boots up in multi worker setup, the above mentioned scenario will go unaddressed as worker with same id won't boot upon a worker failure.

To address this situation an endpoint is exposed which takes `olderThanSeconds` (long) as input. This input is used to determine the duration (from current time) before which the RUNNING tasks must be marker as `FAILED`.

path : `/api/tasks/handle-orphan`
method: `PUT`
response: `200` on success (no body)


suvodeep-pyne (dismissed):
LGTM. 1 cosmetic comment.

This is a nice utility. Although, we need to still solve the underlying problem.

PR #585, opened 2022-07-12
[kubernetes] CronJob for MySQL backup into S3 bucket

shounakmk219:
`CronJob` is a kubernetes resource that schedules a kubernetes `Job` at provided intervals.
This PR aims to 

1. create a `CronJob` which will take a MySQL dump from the ThirdEye MySQL server
2. push that dump to a S3 bucket. 

The cron job can be enabled and configured as below
```
mysql:
  backup:
    enabled: false
    name: backup-folder-name
    schedule: "0 2 * * *"
    s3:
      bucket: thirdeye-backup
      region: us-west-2
```

```
secrets:
  S3BackUpId:
    env: AWS_ACCESS_KEY_ID
    value: <access key id>
  S3BackUpKey:
    env: AWS_SECRET_ACCESS_KEY
    value: <access key>
```

- Backup cron job is enabled/disabled by the `mysql.backup.enabled` flag
- The backup dump job interval is decided by `mysql.backup.schedule` cron expression
- For S3 operation the 2 secrets are required `S3BackUpId` and `S3BackUpKey`


cyrilou242:
Backup LGTM.

Could you add a recovery script?  So that if we have to recover a backup: 
- we have a recovery script that works
- we can do it quickly. 

In the form of a k8s Job could be interesting to reuse the same configuration. 
If it's complicated a shell script should be good to start with.

suvodeep-pyne (approved):
LGTM. This is great work!

Just 1 nit: I want to make the secret name a bit more explicit in case. 
- `mysqlBackupAwsKeyId` instead of `S3BackUpId`
- `mysqlBackupAwsAccessKey` instead of `S3BackUpKey`

PR #589, opened 2022-07-15
major(ui): use platform-ui package instead platform folder

chirag-madlani:
#### Changes:

- Remove `src/app/platform` folder from ThirdEye
- Update references from `platform/components` to `@startree-ui/platform-ui`


jayeshchoudhary:
not required anymore

PR #592, opened 2022-07-19
[server] Alert Performance metrics

shounakmk219:
Exposing below metrics:
- `anomalyPrecision`
- `anomalyResponseRate`

These metrics will be cached for `1 hour` as they are not time sensitive.


thirdeye-server/src/main/java/ai/startree/thirdeye/core/ConfusionMatrix.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.core;
+
+public class ConfusionMatrix {
```
cyrilou242, line 16: MINOR. For metrics classes I've often seen implementations that are more fluent-api-like, and that requires less knowledge by the calling class. 

Eg
ConfusionMatrix
public addTruePositive()
public addTrueNegative()
public addFalsePositive()
public addFalseNegative()
public addUnclassified()
+ public addXXX(int n) to add many observations at the same time

and the ConfusionMatrix maintains the count internally. AppAnalyticsService does not have to know how to count.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
```diff
@@ -79,4 +100,32 @@ private MonitoredMetricWrapper wrapMonitoredMetric(final AlertMetadataDTO metada
         .setDataset(metadata.getDataset().getName())
         .setDatasource(metadata.getDatasource().getName());
   }
+
+  public ConfusionMatrix computeConfusionMatrixForAnomalies() {
+    int truePositive = 0;
+    int falsePositive = 0;
+    int unclassified;
+    List<MergedAnomalyResultDTO> anomalies = anomalyManager.findAll();
```
cyrilou242, line 108: I think child anomalies should not be counted. I think there is no opportunity for the user to give feedback on them.

cyrilou242, line 108: the query reads all the anomalies, is this on purpose?

anomalies without feedback can be obtained with 
```
count(*)  from merged_anomaly_result_index where anomaly_feedback_id = 0
```
--


relevant feedback ids can be obtained with
```
select anomaly_feedback_id from merged_anomaly_result_index where anomaly_feedback_id != 0
```
--> then you can fetch the feedbacks only, rather than the whole anomalies

Understand it's more work though. @suvodeep-pyne may have an opinion.

cyrilou242 (changes requested):

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
cyrilou242: understand they are filtered after but filtering at the db level will have better performance

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
```diff
@@ -79,4 +100,32 @@ private MonitoredMetricWrapper wrapMonitoredMetric(final AlertMetadataDTO metada
...
+    unclassified = anomalies.size() - feedbackTypes.size();
+    for (AnomalyFeedbackType type : feedbackTypes) {
+      switch (type) {
+        case NO_FEEDBACK:
+          unclassified++;
+          break;
+        case NOT_ANOMALY:
+          falsePositive++;
+          break;
+        default:
```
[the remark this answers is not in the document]
shounakmk219: Makes sense. Added separate handling of FeedbackTypes.
Not throwing an exception for invalid type, just logging an error for it as it may break the metric calculation.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/ConfusionMatrix.java
shounakmk219: Got it. Created methods to `inc`, `dec` and `add` to the variables.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
shounakmk219: Make sense. Done.

shounakmk219: Exposed new APIs from MergedAnomalyResultManager to accommodate this usecase.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/ConfusionMatrix.java
```diff
@@ -0,0 +1,123 @@
...
+
+  public void incFalseNegative() {
+    this.falseNegative++;
+  }
+
+  public void incUnclassified() {
+    this.unclassified++;
+  }
+
+  public void decTruePositive() {
```
cyrilou242, line 88: not sure the dec will ever be used - you had a use case in mind?

cyrilou242 (approved):

cyrilou242:
LGTM!

thirdeye-server/src/main/java/ai/startree/thirdeye/core/ConfusionMatrix.java
shounakmk219: Not anything in particular.

PR #606, opened 2022-07-27
[core] Delay operator to introduce delay in pipeline execution.

shounakmk219:
Operator params:

- `delayTime` : delay time in milliseconds

The operator will add a delay to the pipeline using `sleep()`.
Any input passed to it will be carry forwarded as output from the operator


thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class DelayOperator extends DetectionPipelineOperator{
+
+  private static final String DELAY_TIME = "delayTime";
+  private static final String RANDOM_OFFSET = "randomOffset";
+  @Override
+  public void execute() throws Exception {
+    long delayTime = Long.parseLong(getPlanNode().getParams().get(DELAY_TIME).toString());
```
suvodeep-pyne, line 27: Let's use `Duration` instead of `long`. We use `Duration` for all time based inputs in configs or alerts.

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
```diff
@@ -0,0 +1,40 @@
...
+    final AtomicInteger randomOffset = new AtomicInteger();
+    optional(getPlanNode().getParams().get(RANDOM_OFFSET))
```
suvodeep-pyne, line 29: I don't quite understand this random offset thing. What is the requirement here?

shounakmk219: The requirement is to provide a way to avoid having same task execution time for each execution and have something close to a real life behaviour.

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
shounakmk219: Done

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
suvodeep-pyne: As discussed, let's remove the random offset feature.

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class DelayOperator extends DetectionPipelineOperator{
+
+  private static final String DELAY_TIME = "delayTime";
+  private static final String RANDOM_OFFSET = "randomOffset";
+  @Override
+  public void execute() throws Exception {
+    Duration delayTime = Duration.parse(getPlanNode().getParams().get(DELAY_TIME).toString());
```
suvodeep-pyne, line 28: Let's check with @cyrilou242 if there are any standard utils for doing this. There is `TimeUtils` I think which can help here.

suvodeep-pyne (approved):
LGTM. Please address comments and merge.

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
cyrilou242: There are some utils for ISO-8601 Periods (P1D, P7D, etc...). 
If we plan sleeps of 1 day or more then we should use Periods to avoid timezones issues, but I understand this is not required.

In this case I'm fine with durations. I don't think there are utils for this. Though I think jackson is able to parse millis as duration automatically. So if a spec is introduced with typed params then the parsing should be done automatically. 
(not sure though, to test, and this makes the code heavier)

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
shounakmk219: done

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/DelayOperator.java
shounakmk219: Yeah seconds to Duration parsing is taken care by jackson, but as in other operators the time duration input is in the ISO-8601 format hence kept the same for consistency.

PR #611, opened 2022-07-27
feat(ui): support slack and webhook when reviewing in wizard

spham92:
Support showing slack and webhook as well as new sendgrid email while also supporting legacy:

image

### DIFF
--- a/thirdeye-ui/src/app/components/subscription-group-wizard/groups-editor/groups-editor.interfaces.ts
+++ b/thirdeye-ui/src/app/components/subscription-group-wizard/groups-editor/groups-editor.interfaces.ts
@@ -34,4 +34,6 @@ export interface SpecUIConfig {
      */
     /* eslint-disable  @typescript-eslint/no-explicit-any */
     formComponent: FunctionComponent<any>;
+    /* eslint-disable  @typescript-eslint/no-explicit-any */
+    reviewComponent: FunctionComponent<any>;
 }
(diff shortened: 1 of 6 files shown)


thirdeye-ui/src/app/components/subscription-group-wizard/groups-editor/groups-editor.interfaces.ts
```diff
@@ -34,4 +34,6 @@ export interface SpecUIConfig {
      */
     /* eslint-disable  @typescript-eslint/no-explicit-any */
     formComponent: FunctionComponent<any>;
+    /* eslint-disable  @typescript-eslint/no-explicit-any */
+    reviewComponent: FunctionComponent<any>;
```
joshigaurava, line 38: Can't we just use `FunctionComponent`?

spham92: removing any will cause compile issues such as:

```
[tsl] ERROR in /Users/spham/projects/thirdeye/thirdeye-ui/src/app/components/subscription-group-wizard/subscription-group-renderer/subscription-group-renderer.component.tsx(139,45)
      TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '{ configuration: NotificationSpec; }' is not assignable to parameter of type 'Attributes'.
      Object literal may only specify known properties, and 'configuration' does not exist in type 'Attributes'.
```

```
ERROR in /Users/spham/projects/thirdeye/thirdeye-ui/src/app/components/subscription-group-wizard/groups-editor/groups-editor.utils.ts
./src/app/components/subscription-group-wizard/groups-editor/groups-editor.utils.ts 41:42-55
[tsl] ERROR in /Users/spham/projects/thirdeye/thirdeye-ui/src/app/components/subscription-group-wizard/groups-editor/groups-editor.utils.ts(41,43)
      TS2339: Property 'configuration' does not exist on type '{ children?: ReactNode; }'.
```

PR #621, opened 2022-07-29
[security] remove h2 dependency

cyrilou242:


cyrilou242:
let the h2 legacy logic for reference 
we can remove it later

PR #631, opened 2022-08-02
[core] make node params templatable

cyrilou242:
params was `Map<String, Object>`
to allow it to be templatable, `Map<String, Templatable<Object>>` is necessary. 

Using a `Map<String, Templatable<Object>>` required changes in many places and was making code hard to read, so a 
`TemplatableMap<String, Object> extends HashMap<String, Templatable<Object>>` is introduced. 
It makes it simpler to manipulate the map of templatable objects.

Main changes: 
- [TemplatableMap](https://github.com/startreedata/thirdeye/pull/631/files#diff-00a9f19cf03ecdd44ca466cde6fbd6046b1c4a59abb6b6d138fe3745c4ac370c)
- PlanNodeBean and PlanNodeApi

TODO: 
- [ ] do changes in startree-thirdeye

### DIFF
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java
@@ -136,7 +136,7 @@ public class AlertTemplateRenderer {
           // TODO spyne remove magic string. This was done to remove dependency of AnomalyDetector.TYPE on the renderer
           .filter(node -> node.getType().equals("AnomalyDetector"))
           .forEach(node -> node.getParams()
-              .put("anomaly.source", String.format("%s/%s", alertName, node.getName())));
+              .putValue("anomaly.source", String.format("%s/%s", alertName, node.getName())));
     }
 
     return StringTemplateUtils.applyContext(template, properties);
--- a/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/CombinerOperator.java
+++ b/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/CombinerOperator.java
@@ -18,6 +18,7 @@ import static java.util.Collections.emptyMap;
 import static java.util.Objects.requireNonNull;
 
 import ai.startree.thirdeye.detectionpipeline.operator.ForkJoinOperator.ForkJoinResult;
+import ai.startree.thirdeye.spi.datalayer.TemplatableMap;
 import ai.startree.thirdeye.spi.detection.model.DetectionResult;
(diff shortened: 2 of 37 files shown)


thirdeye-spi/src/test/java/ai/startree/thirdeye/spi/api/PlanNodeApiTest.java
```diff
@@ -26,7 +29,7 @@ public class PlanNodeApiTest {
   @Test
   public void testInputNode() throws IOException {
     URL resource = PlanNodeApiTest.class.getClassLoader().getResource("inputNode.json");
-    final PlanNodeApi inputNode = new ObjectMapper().readValue(
+    final PlanNodeApi inputNode = ThirdEyeSerialization.newObjectMapper().readValue(
```
suvodeep-pyne, line 32: Seems like `ThirdEyeSerialization.newObjectMapper()` should be injected. Previously this was `static` and it made sense. but now it seems we have a defined way of conducting this everywhere.

thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java
```diff
@@ -136,7 +136,7 @@ private AlertTemplateDTO applyContext(final AlertTemplateDTO template,
           // TODO spyne remove magic string. This was done to remove dependency of AnomalyDetector.TYPE on the renderer
           .filter(node -> node.getType().equals("AnomalyDetector"))
           .forEach(node -> node.getParams()
-              .put("anomaly.source", String.format("%s/%s", alertName, node.getName())));
+              .putValue("anomaly.source", String.format("%s/%s", alertName, node.getName())));
```
suvodeep-pyne, line 139: Why not use override `put` ?

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/operator/CombinerOperator.java
```diff
@@ -38,7 +39,7 @@ public CombinerOperator() {
   @Override
   public void init(final OperatorContext context) {
     super.init(context);
-    params = optional(getPlanNode().getParams()).orElse(emptyMap());
+    params = optional(getPlanNode().getParams()).map(TemplatableMap::valueMap).orElse(emptyMap());
```
suvodeep-pyne, line 42: Why not override `get()` instead?

suvodeep-pyne (commented):
Let's chat on this sometime

thirdeye-spi/src/test/java/ai/startree/thirdeye/spi/api/PlanNodeApiTest.java
cyrilou242: Will do in a next PR.

PR #635, opened 2022-08-05
[server] Expose stats related to anomaly and feedback

shounakmk219:
The anomaly feedback related stats are exposed under AppAnalyticsApi as below:
```
{
  ...
  "anomalyStats": {
    "totalCount": 110,
    "countWithFeedback": 20,
    "feedbackStats": {
      "NO_FEEDBACK": 0,
      "ANOMALY": 15,
      "NOT_ANOMALY": 2,
      "ANOMALY_NEW_TREND": 0,
      "ANOMALY_EXPECTED": 3
    }
  }
}
```


thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyStatsApi.java
```diff
@@ -0,0 +1,63 @@
...
+    return feedbackStats;
+  }
+
+  public AnomalyStatsApi setFeedbackStats(
+      final Map<AnomalyFeedbackType, Long> feedbackStats) {
+    this.feedbackStats = feedbackStats;
+    return this;
+  }
+
+  public void incFeedbackStatCount(AnomalyFeedbackType type) {
```
suvodeep-pyne, line 60: Just use a setter here. Again, trying to stick to a convention where data pojos are simple pojos with no fancy operations.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
```diff
@@ -124,4 +127,19 @@ public ConfusionMatrix computeConfusionMatrixForAnomalies() {
...
+        .map(MergedAnomalyResultDTO::getFeedback)
+        .collect(Collectors.toList());
+  }
+
+  public AnomalyStatsApi computeAnomalyStats() {
+    List<AnomalyFeedback> feedbacks = getAnomalyFeedbacks();
+    AnomalyStatsApi stat = new AnomalyStatsApi()
+        .setTotalCount(anomalyManager.countParentAnomalies())
+        .setCountWithFeedback((long) feedbacks.size());
+    feedbacks.forEach(feedback -> stat.incFeedbackStatCount(feedback.getFeedbackType()));
```
suvodeep-pyne, line 142: I would just build the map here and set it on the POJO. so we don't have to even initialize it.

Motivation here is to keep business logic classes and data classes separate. so data classes like API classes are completely dumb! only job is to move around the resource classes and get serialized to json.

suvodeep-pyne (approved):
LGTM. minor comments. please address and merge.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
shounakmk219: Makes sense. Done.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyStatsApi.java
shounakmk219: Got it. Done.

PR #651, opened 2022-08-23
[server] add alert insights endpoint

cyrilou242:
Add route to get alert insights: 
- dataset start/end time, 
- recommended default timeframe for the timeseries chart
- others like timezone, granularity, etc...

`GET /alerts/{id}/insights`
returns
```
{
  "templateWithProperties": [AlertTemplateApi]   # contains: timezone, granularity, nodes with properties applied, ...
  "datasetStartTime": 12345678,
  "datasetEndTime": 56789000,
  "defaultStartTime": 12345678,  # recommended default timeframe start for chart
  "defaultEndTime": 56789000,   # recommended default timeframe end for chart
}
```

cc @spham92 

Default timeframe is computed with a simple rule of thumb for the moment. 
Eg: if granularity is 15 minutes --> timeframe size is 1 month
If the granularity is 1 day --> timeframe size is 1 year. 
This should avoid evaluation calls on big timeseries when the granularity is small.


spham92:
response format lgtm

PR #659, opened 2022-08-26
[spi] refactor DetectionPipelineResult

cyrilou242:
- `DetectionPipelineResult` has only one method `List<DetectionResult> getDetectionResults()`
- `DetectionResult` is an interface that contains all the previous methods: `getAnomalies`, `getTimeseries`, etc...
Most classes that implemented DetectionPipelineResult now implement `DetectionResult`
- the implementation that contains anomaly result is `AnomalyDetectionResult`

Impact:
- easier to understand detectionPipelineResult
- should make it easy to have a debug mode, by moving important Datatable methods down to `DetectionResult`. 
- dimensions exploration anomalyMerge is almost implemented. Anomalies are sent to the AnomalyMerger grouped by enumerationItem. 

Notes: 
Following changes were not made because it was adding too much diff to the PR
- better names for classes described above
- reduce for loops, change loop style to stream style
- remove `WrappedAnomalyDetectionResult`, replace by a setter in  `AnomalyDetectionResult`
- anomaly enrichment happens in `DetectionPipelineRunner` and in `AnomalyDetectorOperator`. Everything could be moved to `DetectionPipelineRunner`, reducing `AnomalyDetectorOperator` boilerplating.
- make `DetectionPipelineResult` a class ? no reason it should be an interface anymore
- refactor `getDetectionResults().get(0)` by `getSingleResult()` to make the assumption of single result explicit 

Risks: 
- some places had check `... instanceof DetectionPipelineResult` or `... instanceof DataTable` - need to make sure all theses places are updated correctly.


cyrilou242:
merging and will remove the DetectionPipelineResult entirely in a next PR
this will require to add temporary casting and logic for the CombinerResult somewhere

PR #676, opened 2022-09-08
[persistence] Pulled out db operations to a separate service

shounakmk219:


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/database/DatabaseTaskService.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import java.sql.Connection;
+import java.sql.SQLException;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Singleton
+public class DatabaseTaskService {
```
suvodeep-pyne, line 14: Don't quite understand this. What is the main layer of separation here?

suvodeep-pyne:
@shounakmk219 
This looks great! IMO let's not even think about splitting things into `DatabaseTaskService`. I think we need to rethink that part separately.

right now, the dependency looks like 
```
GenericPojoDao > DatabaseService > DatabaseTaskService
```

And this structure is not clear to me.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/database/DatabaseTaskService.java
shounakmk219: I was thinking the `DatabaseService` would decide which db operation to perform and let `DatabaseTaskService` worry about the transactional aspect of executing the operations and handling the connection etc.

shounakmk219:
> And this structure is not clear to me.

Fair enough, will merge them into single `DatabaseService`

cyrilou242:
Awesome work! 

these methods have a lot of duplicated code
```
public <E extends AbstractIndexEntity> Long count(Class<E> clazz)
public <E extends AbstractIndexEntity> Long count(Predicate predicate, Class<E> clazz)
```

```
public Integer deleteByBaseId(final List<Long> idsToDelete, final Class<? extends AbstractIndexEntity> entityClass)
public Integer delete(final List<Long> idsToDelete, final Class<? extends AbstractEntity> entityClass)
```

```
public <E extends AbstractEntity> List<E> findAll(final Class<E> clazz)
public <E extends AbstractEntity> List<E> findAll(final Predicate predicate, final Long limit, final Long offset, final Class<E> clazz)
```

Maybe they could be refactored to only have one method implementing the logic, and the other method calling it. Similar to how it's done for 
```
public <E extends AbstractEntity> Integer update(final E entity)
public <E extends AbstractEntity> Integer update(final E entity, final Predicate predicate) 
```

shounakmk219:
> these methods have a lot of duplicated code

Yes that's true. There needs to be some refactor on `SqlQueryBuilder` before we can refactor them to have the logic at one place.
Will handle that in a separate PR.

cyrilou242:
I understand 

```
Predicate.EQ("id", id),
```
can be removed from `TaskManagerImpl$updateStatusAndWorkerId`

shounakmk219:
That's correct. Removed the predicate.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/database/DatabaseTaskService.java
suvodeep-pyne: Nice thought here. However, let's tackle this in a separate PR

suvodeep-pyne (approved):
LGTM. Great stuff!

This makes the next piece of work much simpler

cyrilou242:
LGTM! Let's cut a release just before this is merged just in case.

Let's fix the createTime/updateTime currentTimeMillis() quickly after. 
As of today they createTime and updateTime can have different values even if they come from the same insert transaction.

cyrilou242 (approved):

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
```diff
@@ -115,70 +76,49 @@ public List<String> getIndexedColumns(final Class beanClass) {
...
+      return null;
     }
   }
 
+  private <E extends AbstractDTO> GenericJsonEntity toGenericJsonEntity(final E pojo)
+      throws JsonProcessingException {
+    GenericJsonEntity ret = new GenericJsonEntity();
+    int version = pojo.getVersion() == 0 ? 1 : pojo.getVersion();
+    ret.setId(pojo.getId());
+    ret.setCreateTime(new Timestamp(System.currentTimeMillis()));
```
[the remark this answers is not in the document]
shounakmk219: Will note this for a separate PR along with test cases to check they have same values.

PR #678, opened 2022-09-09
[spi] add anomaly labels

cyrilou242:
See design doc: 
https://docs.google.com/document/d/1YVsTY_c0RFZBKYz757M9aEboYB5kixppgRvqbJmsZ04/edit#

This is the implementation of Step 1.
It introduces anomaly labels. 
Labels can be used to filter anomalies. 
Eg: filter anomalies that have the label “cold-start”.

### DIFF
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/mapper/AnomalyLabelMapper.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/mapper/AnomalyLabelMapper.java
@@ -1,3 +1,16 @@
+/*
+ * Copyright 2022 StarTree Inc
+ *
+ * Licensed under the StarTree Community License (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at http://www.startree.ai/legal/startree-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package ai.startree.thirdeye.mapper;
 
 import ai.startree.thirdeye.spi.api.AnomalyLabelApi;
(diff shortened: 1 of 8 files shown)


suvodeep-pyne (approved):
LGTM

Question: How are we planning to filter out anomalies based on labels? Plain text search might be expensive.

cyrilou242:
> LGTM
> 
> Question: How are we planning to filter out anomalies based on labels? Plain text search might be expensive.

@suvodeep-pyne 
To know if an anomaly should be ignored, we can look at the `ignore` field of the labels. 
If one of the label has ignore=true, then the anomaly should be ignored by default. 
On backend side, I don't really plan filtering out based on label.

No changes on the anomaly resource side for the moment. We don't provide backend filtering based on labels.
The reason is that `labeled anomaly` is a subset of  `all anomalies` and to my knowledge we did not observe performance issues when returning all anomalies on a timeframe. 


### Alert page flow: 
Flow on the UI side:
UI makes a single call - no change.
By default UI does not show anomalies with a label ignore=true
UI has a checkbox: show ignored anomalies. ignored anomalies can be shown in a different color/pattern.

For a first version labels are shown in the infobox on hover. 
exact UI to be defined with Edwin/Madhu/Steven

I'm planning name-based filtering on the UI side only for the moment. 
Instead of "show ignored anomalies" buttong, we can have the names of labels. 
Eg: 

First version
- [ ] Show Ignored anomalies

Second version
Show Ignored anomalies
- [ ] Christmas
- [X] Cold start
- [ ] Change ratio limit

### Anomalies page
similar - UI fetches all anomalies on the timeframe as usual
display of anomalies with labels ignore=true to be defined

Does this help?

suvodeep-pyne:
Got it. 

In essense, these nodes are pass through nodes. meaning whatever comes in, goes out. Can I use this with dimension exploration out of the box?

cyrilou242:
> Got it.
> 
> In essence, these nodes are pass through nodes. meaning whatever comes in, goes out. Can I use this with dimension exploration out of the box?

yes exactly @suvodeep-pyne 

For dimension exploration, the goal is to have the PostProcessors out of the dimension exploration block.
Simple postProcessors like cold start and event will only use `getAnomalies()`, so calling getAnomalies() of `CombinerResult` will be fine.
More advanced postProcessor like deviation-threshold need to look at the data. I'll try to have the CombinerResult handling logic in the Operator to hide it from the PostProcessor implementations. 

wip PostProcessor classes are in the second PR: 
https://github.com/startreedata/thirdeye/pull/686/files#diff-dd9f0d3c999e0a424c348c14f2f8396d8b2303bd5f9eb00963daf7b9c95585e4

https://github.com/startreedata/thirdeye/pull/686/files#diff-6c2571138246f96e5f4a997b816b4d09cd8c71dcd70ee4926295431dda8ec55b

PR #688, opened 2022-09-16
[k8s] make helm template javaOpts compatible with commas

cyrilou242:
some java options can have commas in their arguments: 

```
-XX:StartFlightRecording=settings=default,delay=3m,disk=true,name=te-run,filename=te-run.jfr,maxsize=500M,maxage=2h
```


shounakmk219 (approved):
LGTM

PR #698, opened 2022-09-27
[persistence] Separate table for task entities

shounakmk219:
- `TaskManagerImpl` no more extends `AbstractManagerImpl` . As a result `TaskManagerImpl` now implements all the methods from `TaskManager`+`AbstractManager`
- `TaskManagerImpl` now has `TaskDao` as constructor param instead of `GenericPojoDao`
- New `TaskDao` supports all the db operation flows that `GenericPojoDao` used to support for `TaskManagerImpl`
- New bean class `TaskEntity` 
- DB migration file to create a new table `task_entity` and its required indexes


suvodeep-pyne:
Hey @shounakmk219 

Good work on this. Let's roll on this idea but split it into different PRs again.
- Step/PR 1: Let's add implementation for `Task`, `TaskDao` and tests `TestTestDao`. Let's not wire this in yet. Meaning, the TaskManager and everything underneath remains unchanged.
- Step/PR 2: At this stage we have a new `Task` entity which we know works and sits in the codebase. Step 2 should be to wire this up.

I'll also add some minor comments on the PR.

suvodeep-pyne (changes requested):

suvodeep-pyne:
How do we generate id for the new Task Table?

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_75_0__add_task_entity_table.sql
```diff
@@ -0,0 +1,25 @@
...
+    end_time    bigint(20)   NOT NULL,
+    job_id      bigint(20),
+    worker_id   bigint(20),
+    create_time timestamp,
+    update_time timestamp DEFAULT CURRENT_TIMESTAMP,
+    version     int(10),
+    last_active timestamp DEFAULT CURRENT_TIMESTAMP,
+    json_val    text
+    ) ENGINE = InnoDB;
+CREATE INDEX task_status_idx ON task_entity (status);
```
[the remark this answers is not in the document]
shounakmk219: done

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/util/SqlQueryBuilder.java
```diff
@@ -237,7 +237,7 @@ public PreparedStatement createFindByParamsStatementWithLimit(final Connection c
   }
 
   public PreparedStatement createCountStatement(final Connection connection, final Predicate predicate,
-      final Class<? extends AbstractIndexEntity> indexEntityClass) throws Exception {
+      final Class<? extends AbstractEntity> indexEntityClass) throws Exception {
```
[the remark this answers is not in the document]
shounakmk219: Actually it's the other way round, all the createStatements are accepting a class that extends AbstractEntity. Also the place where `tableToEntityNameMap` is populated accepts class that extends AbstractEntity. So passing a class that extends AbstractIndexEntity to get value from `tableToEntityNameMap` map seem like unnecessary restriction. Let me know if I am missing something here

shounakmk219:
> How do we generate id for the new Task Table?

We don't for the task_entity table. It is a standalone table and does not depend on any base table hence having unique id just across this table is enough.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/mapper/TaskEntityMapper.java
```diff
@@ -0,0 +1,33 @@
...
+public interface TaskEntityMapper {
+  TaskEntityMapper INSTANCE = Mappers.getMapper(TaskEntityMapper.class);
+
+  @Mapping(source = "jobName", target = "name")
+  @Mapping(source = "taskType", target = "type")
+  @Mapping(target = "jsonVal", ignore = true)
+  @Mapping(target = "createTime", expression = "java(new java.sql.Timestamp(System.currentTimeMillis()))")
+  @Mapping(target = "updateTime", expression = "java(new java.sql.Timestamp(System.currentTimeMillis()))")
+  @Mapping(target = "version", expression = "java(dto.getVersion() == 0 ? 1 : dto.getVersion())")
+  TaskEntity toTaskEntity(TaskDTO dto);
```
suvodeep-pyne, line 32: How's the reverse side of things handled?

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/mapper/TaskEntityMapper.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface TaskEntityMapper {
+  TaskEntityMapper INSTANCE = Mappers.getMapper(TaskEntityMapper.class);
+
+  @Mapping(source = "jobName", target = "name")
+  @Mapping(source = "taskType", target = "type")
+  @Mapping(target = "jsonVal", ignore = true)
+  @Mapping(target = "createTime", expression = "java(new java.sql.Timestamp(System.currentTimeMillis()))")
```
suvodeep-pyne, line 29: Don't want code in quotes. let's find a better and simpler way. nice hack though!

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/mapper/TaskEntityMapper.java
shounakmk219: We are using `OBJECT_MAPPER` for that inside `TaskDao.toDto` as it needs to construct the dto from json value.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/mapper/TaskEntityMapper.java
shounakmk219: Cleaned the expression based mapping👍

suvodeep-pyne (approved):
LGTM

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
```diff
@@ -0,0 +1,315 @@
...
+
+  private TaskEntity toEntity(final TaskDTO dto)
+      throws JsonProcessingException {
+    TaskEntity entity = TaskEntityMapper.INSTANCE.toTaskEntity(dto);
+    final String jsonVal = toJsonString(dto);
+    entity.setJsonVal(jsonVal);
+    final Timestamp currentTime = new java.sql.Timestamp(System.currentTimeMillis());
+    entity.setCreateTime(currentTime);
+    entity.setUpdateTime(currentTime);
+    entity.setVersion(dto.getVersion() == 0 ? 1 : dto.getVersion());
```
suvodeep-pyne, line 67: What's with this weird logic? so 1 is the final version?

let's not make any changes in this PR. but let's have a JIRA for this

shounakmk219: Yes that's a good point. The logic is `toEntity` gets called from `put` and `update`, in `put` flow the version needs to be set as 1 (first version) and for `update` take the version provided by `dto`. `put` operations should take the responsibility of setting the version as 1 before calling `toEntity`. Tried to mimic the logic from `GenericPojoDao.toGenericJsonEntity`. Will handle this separately with some tests to make sure there is nothing more to it.

PR #713, opened 2022-10-05
chore(ui): TE-1114 add header to not cache index.html

spham92:
Add no cache header and 1hr expiration

### DIFF
--- a/thirdeye-ui/docker/templates/default.conf.template
+++ b/thirdeye-ui/docker/templates/default.conf.template
@@ -28,6 +28,11 @@ server {
         proxy_read_timeout 300;
     }
     location / {
-        try_files $uri $uri/ /index.html;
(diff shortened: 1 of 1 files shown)


thirdeye-ui/docker/templates/default.conf.template
```diff
@@ -29,5 +29,7 @@ server {
     }
     location / {
         try_files $uri $uri/ /index.html;
+        expires 1h;
+        add_header Cache-Control "no-cache ";
```
suvodeep-pyne, line 33: Will this disable cache for everything. I also found this:
https://stackoverflow.com/questions/41631399/disable-caching-of-a-single-file-with-try-files-directive

spham92: Made changes based on the stackoverflow

PR #718, opened 2022-10-11
[core] fix calcite request dimension identifier quoting

cyrilou242:
quickfixes

need proper refactoring


cyrilou242:
will refactor later

PR #725, opened 2022-10-17
[persistence] DatabaseTransactionService to handle transaction and connection

shounakmk219:
## DatabaseTransactionService
A transactional layer between DatabaseService and DAO layer to enable DAO layer for performing multiple DB operations as part of a single transaction. 

image

### Major changes in the PR

- DatabaseService now does not have any GenericPojoDao specific methods.
- DAO layer now depends on DatabaseTransactionService instead of DatabaseService
- The connection handling is moved out from DatabaseService to DatabaseTransactionService which means all the DatabaseService methods will need a connection to be passed.

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/DatabaseService.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/DatabaseService.java
@@ -13,17 +13,11 @@
  */
 package ai.startree.thirdeye.datalayer;
 
-import static ai.startree.thirdeye.datalayer.mapper.DtoIndexMapper.toAbstractIndexEntity;
-
 import ai.startree.thirdeye.datalayer.entity.AbstractEntity;
 import ai.startree.thirdeye.datalayer.entity.AbstractIndexEntity;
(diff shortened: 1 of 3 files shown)


suvodeep-pyne (changes requested):
@shounakmk219 
couple of comments here:
- I think `DatabaseService` should depend on `DatabaseTransactionService`
- Also, therefore DatabaseService doesn't need to take `Connection` as an argument to every function. 
- This makes a straightforward hierarchy. `GenericPojoDao > DatabaseService > DatabaseTransactionService > DataSource`.

suvodeep-pyne (approved):
LGTM.

shounakmk219:
> @shounakmk219 couple of comments here:
> 
> * I think `DatabaseService` should depend on `DatabaseTransactionService`
> * Also, therefore DatabaseService doesn't need to take `Connection` as an argument to every function.
> * This makes a straightforward hierarchy. `GenericPojoDao > DatabaseService > DatabaseTransactionService > DataSource`.

Hey @suvodeep-pyne the hierarchy that you mentioned, that is something that current master has (instead of the transaction service it has the runTask method)
The problem with that was connection was getting committed after each DatabaseService method run, hence making it difficult to create a transaction.

PR #735, opened 2022-10-25
mac m1s use arm64 pinot container

jacksonargo:
Use the arm64 pinot container when running unit tests on an M1 laptop.

### DIFF
--- a/pinot-test-container/src/main/java/org/apache/pinot/testcontainer/PinotContainer.java
+++ b/pinot-test-container/src/main/java/org/apache/pinot/testcontainer/PinotContainer.java
@@ -13,7 +13,9 @@
  */
 package org.apache.pinot.testcontainer;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.time.Duration;
 import java.util.List;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


cyrilou242 (changes requested):
LGTM
not easy to be the first developer running the project on a M1! 

- let's move reader.close() to a `finally` close. this ensures the reader is closed event if readLine fails
```
try {
...
} catch {
...
}finally {
reader.close()
}
```

- L74 let's not catch silently. you could log an error or throw. I'm in favor of throwing and letting the next developer fix the issue, it's unlikely we want to pursue with an incorrect image.  

few java conventions: 
method names start with lowercase:
`ImageTag()` `HostArch()` --> `imageTag()`, ...

in this codebase we try to put `final` when a variable is effectively final. 
this makes the code more robust to changes. 

`final` prevents [the reference hold by an object] / [the value of a primitive] to be changed

jacksonargo:
* I renamed the methods w/ lowercase.
* Catch block throws a runtime exception. My original thought was to let it try the x86_64 image if the systctl command fails, but ya it _shouldn't_ fail, and if it does, it's prob best to surface it.
* Since the catch block is throwing a runtime exception, I left the reader.Close() as is.

pinot-test-container/src/main/java/org/apache/pinot/testcontainer/PinotContainer.java
```diff
@@ -139,7 +172,10 @@ public void addTables() throws IOException, InterruptedException {
 
   private long getTableSize(final String tableName) throws IOException, InterruptedException {
     final ExecResult checkIngestionIsFinished = this.execInContainer(
-        "curl", "-X", "GET", String.format("http://localhost:9000/tables/%s_OFFLINE/size?detailed=false", tableName)
+        "curl",
```
cyrilou242, line 175: it seems we don't have the same code style configuration.
let's loop with Suvodeep and Shounak to ensure everybody has the same.

cyrilou242: not a blocker of course

cyrilou242 (approved):

cyrilou242:
lgtm

pinot-test-container/src/main/java/org/apache/pinot/testcontainer/PinotContainer.java
jacksonargo: 🤷 I imported the styles from https://github.com/startreedata/thirdeye/blob/master/thirdeye-checkstyle.xml, but ya Intellij still didn't like this line.

PR #737, opened 2022-11-01
[server] endpoint for dashboard metrics

shounakmk219:
Endpoint for dashboard related metrics

endpoint : `/api/dashboard?from=<from epoch time>&to=<to epoch time>`
response :
```
{
  "anomalyCount": 150,
  "anomalyCountWithoutFeedback": 149,
  "activeAlertsCount": 3,
  "precision": 1,
  "sgCount": 1,
  "anomalyTs": [
    1661868011667,
    1663667122786,
    1663667122786,
    1663667122787,
    1664174009007
  ]
}
```


suvodeep-pyne (changes requested):
Let's wait to add a `DashboardResource`. So far, we are just pushing counts and recent anomalies, most of which is already available via APIs today. May be we are missing a `/count` endpoint.

PR #738, opened 2022-11-02
[core] replace groovy by commons-text for string templating

cyrilou242:
replace groovy by apache commons text for template variable replacements
fixes an important memory leak described here: 
https://cortexdata.atlassian.net/browse/TE-1038

Risk: 
- some behavior of groovy templating were not unit tested. I tried to add a few tests.

Other options considered:
- update groovy to fix the leak: did not solve the issue - groovy does not expect a runtime to create many templates
- instead of creating a groovy template every time, maintain a hashmap of created groovy templates. Match te templates to groovy templates.
--> means we have to maintain a cache with TTL ourselves. Don't want to do this. Speed is not really an issue here. (pipeline execution time >> template resolution time)

Note on speed: 
I did a very quick benchmark, commons text seems to be faster.
average templating time with groovy: 15ms
average templating time commons-text: 0.5ms
benchmark is not realistic because: 
- groovy was stressed by its own memory leak 
- the template used was very small (10 chars), TE templates are bigger (10000 chars)
- groovy expects a template to be reused, which is not the case in the current implem
At least it shows we should not worry about this for the moment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -795,9 +795,9 @@
       </dependency>
 
       <dependency>
-        <groupId>org.codehaus.groovy</groupId>
-        <artifactId>groovy-all</artifactId>
-        <version>2.4.21</version>
+        <groupId>org.apache.commons</groupId>
+        <artifactId>commons-text</artifactId>
+        <version>1.10.0</version>
       </dependency>
       <dependency>
         <groupId>org.reflections</groupId>
--- a/thirdeye-core/pom.xml
+++ b/thirdeye-core/pom.xml
@@ -74,8 +74,8 @@
 
     <!-- This is used to replace params in a string template -->
     <dependency>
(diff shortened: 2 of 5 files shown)


shounakmk219 (approved):
Great stuff!
LGTM

suvodeep-pyne:
Hey @cyrilou242 !

the idea looks good to me. But pinot uses the same engine. I wonder if they are seeing this as well?

other than that lgtm

cyrilou242:
> Hey @cyrilou242 !
> 
> the idea looks good to me. But pinot uses the same engine. I wonder if they are seeing this as well?
> 
> other than that lgtm

In Pinot it seems the same groovy logic is only used in Pinot for jobSpec rendering for batch.
So they are not abusing createTemplate like we do in TE
will reach out just in case

PR #748, opened 2022-11-04
Disable license header check for UI project

joshigaurava:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-345

#### Description

UI project is in the middle of some configuration updates. As a result, some files are missing license headers. THis is a temporary fix to avoid build failure.


suvodeep-pyne (approved):
LGTM

PR #775, opened 2022-11-11
feat(ui): TE-1002 added date filter in admin page and set the default filter to last 7 days

alcatraz627:
#### Issue(s)

- [TE-1002](https://cortexdata.atlassian.net/browse/TE-1002)

#### Description

1. Added date filter on the admin page 
2. Setting the date filter to the last 7 days by default

#### Screenshots

Screenshot 2022-11-11 at 17 01 09


thirdeye-ui/src/app/components/admin-page/recent-failures/recent-failures.component.tsx
```diff
@@ -66,6 +114,30 @@ export const RecentFailures: FunctionComponent = () => {
...
+                        isError={false}
+                        isLoading={status === ActionStatus.Working}
+                        loadingState={
+                            <SkeletonV1
+                                animation="pulse"
+                                height={40}
+                                width={425}
+                            />
+                        }
+                    >
```
spham92, line 128: We can forego this nesting the time selector in this component. We should always show the time selector in this page

thirdeye-ui/src/app/components/admin-page/recent-failures/recent-failures.component.tsx
```diff
@@ -18,25 +18,73 @@ import TableCell from "@material-ui/core/TableCell";
...
+    useEffect(() => {
+        // Apply the default filter of the last 7 days if no
+        // prior filter is specified or if the filters are invalid
+        if (!startTime || !endTime || startTime > endTime) {
+            const endRange = Date.now();
+            const startRange = endRange - WEEK_IN_MILLISECONDS;
+
+            onHandleTimeRangeChange(startRange, endRange);
+        }
     }, []);
```
spham92, line 88: Typically I would let route redirection take care of setting the default params but this page is a little unique.

For this case, we can definitely simplify this by having the memoize function set the default if the expected parameters are not in the query params. I recommend moving the logic from lines `useEffect` in 79 - 88 to the `useMemo` function in 40-46.  You can remove the duplicate states in 48 and 49

alcatraz627: Trying to update the query params from `useMemo` does not work, and throws a console warning.


Screenshot 2022-11-14 at 14 53 13


Screenshot 2022-11-14 at 14 51 29

alcatraz627: Should I leave the query param updation on page load for page redirection to take care of later on? The reason for using the `useEffect` initially was to get around this issue.

thirdeye-ui/src/app/components/admin-page/recent-failures/recent-failures.component.tsx
alcatraz627: Removed the loading UI

spham92 (approved):
lgtm

PR #777, opened 2022-11-14
fix(ui): TE-1059 renamed labels and replaced a fixed label with a generic label

alcatraz627:
#### Issue(s)

[TE-1059](https://cortexdata.atlassian.net/browse/TE-1059)

#### Description

- Renamed "View all Anomalies" to "View recent Anomalies"
- Replaced "View all Alerts" label with a generic "View all {{entities}" label

#### Screenshots

image


spham92 (approved):
good catch on the view all alerts

PR #789, opened 2022-11-16
[spi] remove protective copy in SimpleDataTable

cyrilou242:
similar changes required in internal distribution


suvodeep-pyne (approved):
LGTM

PR #796, opened 2022-11-22
[server] Analytics service refactor to accept DaoFilter for computations

shounakmk219:


thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
```diff
@@ -115,51 +121,55 @@ private MonitoredMetricWrapper wrapMonitoredMetric(final AlertMetadataDTO metada
...
+  private Map<AnomalyFeedbackType, Long> aggregateFeedbackTypes(final List<AnomalyFeedback> feedbacks) {
+    final Map<AnomalyFeedbackType, Long> feedbackStats = new HashMap<>();
+    for(final AnomalyFeedbackType type : AnomalyFeedbackType.values()) {
       feedbackStats.put(type, 0L);
     }
     feedbacks.forEach(feedback -> {
-      AnomalyFeedbackType type = feedback.getFeedbackType();
-      long count = feedbackStats.get(type);
+      final AnomalyFeedbackType type = feedback.getFeedbackType();
+      final long count = feedbackStats.get(type);
```
cyrilou242, line 170: Minor (only a suggestio,)
These days I tend to prefer
```
feedbackStats.computeIfAbsent(type, () -> 0)
```

it allows to remove
```
    for(final AnomalyFeedbackType type : AnomalyFeedbackType.values()) {
      feedbackStats.put(type, 0L);
    }
```

cyrilou242 (commented):
LGTM.
If I understand correctly filtering is not exposed yet in this PR?

shounakmk219:
> LGTM. If I understand correctly filtering is not exposed yet in this PR?

Yes that's correct. That will be handled in a separate PR.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/AppAnalyticsService.java
shounakmk219: The for loop also handles initialising each feedback type to 0 so that the missing feedback types will show the 0 count to put out all the possible feedback types with their counts.

cyrilou242: oh good point

cyrilou242 (approved):
LGTM

PR #797, opened 2022-11-22
te 809 add template property metadata

cyrilou242:
Add more metadata to template properties.

In alert templates:
Deprecate: 
```json
"defaultProperties": {
  "varName": value
}
```

introduce: 
```json 
"properties": [
  {
    "name": "varName",
    "description": "Does this and that [markdown](link)"
    "defaultValue": "defaultVal",
    "defaultIsNull": false,
    "min": ...
    "max: ...
    "options":  ...
    "jsonType": ...
  }
],
```

Note:
- breaks the fronted @spham92 because the frontend differentiates default to mandatory variables by looking into `defaultProperties`. Need to also look into propertiesMetadata now
- We can add more fields, like javaType, recommended UI component (eg horizontal range vs numeric input box). Just want to make sure this is compatible with dimension exploration use cases @suvodeep-pyne. In particular, will this make it possible to have a "default value for enumeration"?  

Given it breaks the frontend we will migrate all templates once frontend is fixed.
A partial migration in `startree-threshold-percentile` is introduced for frontend testing.

### DIFF
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java
@@ -162,8 +162,7 @@ public class AlertTemplateRenderer {
     final Map<String, Object> legacyDefaultProperties = optional(
         template.getDefaultProperties()).orElse(new HashMap<>());
     properties.putAll(legacyDefaultProperties);
-    final Map<String, Object> defaultProperties = defaultProperties(
-        template.getPropertiesMetadata());
+    final Map<String, Object> defaultProperties = defaultProperties(template.getProperties());
     properties.putAll(defaultProperties);
 
     if (templateProperties != null) {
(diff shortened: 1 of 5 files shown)


suvodeep-pyne:
I'd just call it "properties" now because it is a list of properties with attributes. suggestions:
 - `properties` instead of propertiesMetadata`
 - `default` instead of `defaultValue`
 - `description`: could be long text then I assume or point to documentation.
 - `exampleValue`: what is the use case? If we have description, we should have examples here. Otherwise if we are providing suggestion, we should have multiple something like `options: []` but you already have `allowedValues`
 - rename `allowedValues` to `options` then? something like `"strict": true` could be on the table later
 - `"min" and "max"`. want to avoid having `value` everywhere. doesn't seem to add a lot of value :P


```
"properties": [
  {
    "name": "varName",
    "description": "Does this and that [markdown](link)", // markdown would be amazing. This should be a long text field.
    "jsonType": ...
    "default": "defaultVal", // prefer is to defaultValue. 
    "defaultIsNull": false, // nice!
    "exampleValue": ... // not sure about this
    "min": ...
    "max": ...
    "options":  []
  }
],

```

This should work fine for literals? How about DTO validations etc?
example: field is a datasource, dataset,

cyrilou242:
Thanks @suvodeep-pyne! Will update accordingly. 

> exampleValue: what is the use case?

it's the placeholder in the UI input box, for values that don't have a defaultValue. I saw this in a discussion we had previously. `placeholder` would be a better naming
If you don't see the use case anymore I'm ok to remove it, agree description is enough.
We can always add it back later if some UX analysis suggests to do so.


> How about DTO validations etc?

I was planning to do this in another PR, this is a bit more involved I need to put more thoughts into this.
Eg We could do validation of DTO, and eg provide options for datasets, but then 
what about metricDTO validation/options ? --> it depends on the dataset.
Also need to think of how to put as less logic as possible on the frontend side.

The main goal of this PR is to make filter postprocessors usable. They are not as of today because the users has to look into the template to understand the meaning of the properties.

cyrilou242:
> default instead of defaultValue
`default` is a reserved keyword, keeping defaultValue

PR #798, opened 2022-11-23
[server] endpoint to get stats for a particular alert

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1069
https://cortexdata.atlassian.net/browse/TE-1103

#### Description

New endpoint on `AlertResource` to allow fetching individual alert stats.
endpoint : `/api/alerts/{alertId}/stats`
response:
```
{
  "totalCount": 100,
  "countWithFeedback": 80,
  "feedbackStats": {
    "NO_FEEDBACK": 10,
    "ANOMALY_EXPECTED": 20,
    "ANOMALY": 30,
    "ANOMALY_NEW_TREND": 10,
    "NOT_ANOMALY": 10
  }
}
```
The endpoint also accepts query params to further filter the sample set of anomalies (only enumerationItem.id, startTime, endTime is supported, anything else will be ignored)
example : `/api/alerts/13658/stats?enumerationItem.id=13660&startTime=1661970600000&endTime=1670344199999`
response:
```
{
   "totalCount":12,
   "countWithFeedback":8,
   "feedbackStats":{
      "ANOMALY_NEW_TREND":1,
      "NOT_ANOMALY":1,
      "NO_FEEDBACK":0,
      "ANOMALY_EXPECTED":4,
      "ANOMALY":2
   }
}
```


thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
```diff
@@ -257,4 +267,26 @@ private long safeEndTime(final @Nullable Long endTime) {
...
   }
+
+  @GET
+  @Timed
+  @Path("{id}/stats")
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response getAnalytics(
+      @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
+      @PathParam("id") final Long id,
+      @Context UriInfo uriInfo
```
suvodeep-pyne, line 278: Use `@QueryParam`. if @PathParam exists, so should QueryParam. no?

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
```diff
@@ -257,4 +267,26 @@ private long safeEndTime(final @Nullable Long endTime) {
...
+  ) {
+    ensureExists(id);
+    final MultivaluedMap<String, String> queryParameters = uriInfo.getQueryParameters();
+    final Predicate alertPredicate = Predicate.EQ("detectionConfigId", id);
+    final DaoFilter filter;
+    if (queryParameters.isEmpty()) {
+      filter = new DaoFilter().setPredicate(alertPredicate);
+    } else {
+      filter = new DaoFilterBuilder(AnomalyResource.API_TO_INDEX_FILTER_MAP).buildFilter(queryParameters);
```
suvodeep-pyne, line 287: This logic is for general query param processing in CrudResource. Can you follow a simpler example: https://github.com/startreedata/thirdeye/blob/master/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java#L187

shounakmk219: Restricted the allowed query params for filtering to only enumerationItem.id, startTime, endTime.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
shounakmk219: Done

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
```diff
@@ -257,4 +266,29 @@ private long safeEndTime(final @Nullable Long endTime) {
...
+    final MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+    // required filter
+    queryParameters.put("alert.id", List.of(String.valueOf(id)));
+
+    // optional filters
+    optional(enumerationId).map(enumId -> queryParameters.put("enumerationItem.id", List.of(enumId)));
+    optional(startTime).map(start -> queryParameters.put("startTime", List.of(start)));
+    optional(endTime).map(end -> queryParameters.put("endTime", List.of(end)));
+
+    final DaoFilter filter = new DaoFilterBuilder(AnomalyResource.API_TO_INDEX_FILTER_MAP).buildFilter(queryParameters);
```
suvodeep-pyne, line 291: Why not provide the predicate directly?

This is pretty roundabout. 
```
QueryParam > Map > DaoFilterBuilder > DaoFilter.
```

Let's simplify QueryParam > DaoFilter.

shounakmk219: Query params with operators such as `startTime=[gte]166197060000` needs to be parsed to get something like Predicate.GE("startTime", 166197060000). Currently this operator map is private to DaoFilterBuilder and the only public method DaoFilterBuilder exposes is the one that takes query params map to build DaoFilter. Let me know if I am missing something here.
Else we may need to refactor the DaoFilterBuilder to skip the roundabout.

shounakmk219: Removed the complexity of having operators in query params. It makes sense to just accept the start and end time without operators. Was thinking about making it consistent to anomaly filtering api call hence had the params in query initially.

suvodeep-pyne (approved):
LGTM.

tip: for optional, if you are not using the value after `map()`, you can use `ifPresent()` to return void.

PR #803, opened 2022-11-28
Access control hooks for GET and DELETE

jacksonargo:
Jira: https://cortexdata.atlassian.net/browse/TE-1096

This change adds access control hooks for getting and deleting ThirdEye generic resources. I'm saving the PUT/POST methods and resource specific methods for subsequent prs.

The hooks check that the requester has access to do whatever operation. The access control check always returns true, so no API behavior is changed.

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
@@ -157,11 +157,11 @@ public class AlertResource extends CrudResource<AlertApi, AlertDTO> {
       @Context java.net.http.HttpHeaders httpHeaders
   ) {
     final AlertDTO dto = get(id);
-    final AlertInsightsApi insights = alertInsightsProvider.getInsights(dto);
     if (!hasAccess(dto, AccessType.READ, httpHeaders)) {
       return Response.status(Status.FORBIDDEN).build();
     }
 
+    final AlertInsightsApi insights = alertInsightsProvider.getInsights(dto);
     return Response.ok(insights).build();
   }
 
@@ -169,10 +169,8 @@ public class AlertResource extends CrudResource<AlertApi, AlertDTO> {
   @POST
   @Timed
   @Produces(MediaType.APPLICATION_JSON)
-  public Response getInsights(
-      @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
-      final AlertInsightsRequestApi request
-  ) {
+  public Response getInsights(@ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
+      final AlertInsightsRequestApi request) {
     final AlertApi alert = request.getAlert();
     ensureExists(alert);
 
@@ -187,13 +185,17 @@ public class AlertResource extends CrudResource<AlertApi, AlertDTO> {
       @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
       @PathParam("id") final Long id,
       @FormParam("start") final Long startTime,
-      @FormParam("end") final Long endTime
+      @FormParam("end") final Long endTime,
+      @Context java.net.http.HttpHeaders httpHeaders
   ) {
+    final AlertDTO dto = get(id);
+    ensureExists(dto);
     ensureExists(startTime, "start");
-    ensureExists(get(id));
+    if (!hasAccess(dto, AccessType.UPDATE, httpHeaders)) {
+      return Response.status(Status.FORBIDDEN).build();
+    }
 
     alertCreater.createOnboardingTask(id, startTime, safeEndTime(endTime));
-
     return Response.ok().build();
   }
 
@@ -243,8 +245,14 @@ public class AlertResource extends CrudResource<AlertApi, AlertDTO> {
   @Produces(MediaType.APPLICATION_JSON)
   public Response reset(
       @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
-      @PathParam("id") final Long id) {
+      @PathParam("id") final Long id,
+      @Context java.net.http.HttpHeaders httpHeaders
+  ) {
     final AlertDTO dto = get(id);
+    if (!hasAccess(dto, AccessType.READ, httpHeaders)) {
+      return Response.status(Status.FORBIDDEN).build();
+    }
+
     LOG.warn(String.format("Resetting alert id: %d by principal: %s", id, principal.getName()));
 
     alertDeleter.deleteAssociatedAnomalies(dto.getId());
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
@@ -31,6 +31,7 @@ import ai.startree.thirdeye.spi.api.AlertEvaluationApi;
 import ai.startree.thirdeye.spi.api.AlertInsightsApi;
 import ai.startree.thirdeye.spi.api.AlertInsightsRequestApi;
 import ai.startree.thirdeye.spi.api.UserApi;
+import ai.startree.thirdeye.spi.authorization.AccessType;
 import ai.startree.thirdeye.spi.datalayer.bao.AlertManager;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
 import com.codahale.metrics.annotation.Timed;
@@ -55,9 +56,11 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
 import org.apache.commons.lang3.StringUtils;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.quartz.CronExpression;
@@ -148,10 +151,16 @@ public class AlertResource extends CrudResource<AlertApi, AlertDTO> {
   @Timed
   @Produces(MediaType.APPLICATION_JSON)
   @Deprecated(forRemoval = true)
-  public Response getInsights(@ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
-      @PathParam("id") final Long id) {
+  public Response getInsights(
+      @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
+      @PathParam("id") final Long id,
+      @Context java.net.http.HttpHeaders httpHeaders
+  ) {
     final AlertDTO dto = get(id);
     final AlertInsightsApi insights = alertInsightsProvider.getInsights(dto);
+    if (!hasAccess(dto, AccessType.READ, httpHeaders)) {
+      return Response.status(Status.FORBIDDEN).build();
+    }
 
     return Response.ok(insights).build();
   }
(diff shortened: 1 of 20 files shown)


thirdeye-server/pom.xml
```diff
@@ -92,6 +92,12 @@
       <artifactId>mysql</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+      <version>2.1</version>
```
suvodeep-pyne, line 98: add the main dependency to the parent pom. and in the server pom, you don't need the version. this allows us to control all dependencies from the parent pom and is a general practice.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
```diff
@@ -148,11 +151,15 @@ protected AlertApi toApi(final AlertDTO dto) {
   @Timed
   @Produces(MediaType.APPLICATION_JSON)
   @Deprecated(forRemoval = true)
-  public Response getInsights(@ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
-      @PathParam("id") final Long id) {
+  public Response getInsights(
+      @ApiParam(hidden = true) @Auth ThirdEyePrincipal principal,
+      @PathParam("id") final Long id,
+      @Context java.net.http.HttpHeaders httpHeaders
```
suvodeep-pyne, line 157: you can probably simplify the qualified import

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/authorization/EntityType.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.spi.authorization;
+
+public enum EntityType {
```
suvodeep-pyne, line 16: What is this entityType for? We already have different entity types in the persistence layer

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/authorization/AccessControl.java
```diff
@@ -0,0 +1,25 @@
...
+package ai.startree.thirdeye.spi.authorization;
+
+import java.net.http.HttpHeaders;
+
+public interface AccessControl {
+
+  boolean hasAccess(
+      AccessControlIdentifiers identifiers,
+      AccessType accessType,
+      HttpHeaders httpHeaders
```
suvodeep-pyne, line 23: Instead of `httpHeaders`, can we structure the exact information that we need in the interface for the implementation to be able to decide on the access control

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/authorization/AccessControlIdentifiers.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.spi.authorization;
+
+public class AccessControlIdentifiers {
```
suvodeep-pyne, line 16: Since it points to a single entity probably should be singular `AccessControlIdentifier`

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AbstractDTO.java
```diff
@@ -90,6 +92,14 @@ public String toString() {
     return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
   }
 
+  public AccessControlIdentifiers toAccessControlIdentifiers() {
```
suvodeep-pyne, line 95: Having this in the persistence layer is a bit tricky. The dtos should be pojos. Let's have an utility in the server for creating an ID from any entity.

suvodeep-pyne (changes requested):
@jacksonargo 
This is great progress. Have a few comments around code organization. Also, would to understand more about how the implementation is gonna work.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
jacksonargo: This file is also importing `javax.ws.rs.core.HttpHeaders`, altho I'm not really sure if there's a difference.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/authorization/EntityType.java
jacksonargo: This is used for the startree resource name https://github.com/startreedata/startree-pinot/blob/master/startree-plugins/startree-auth-platform/src/main/java/ai/startree/pinot/auth/platform/SRN.java

I _kinda_ think it makes sense to use a separate entity type enum for the access control entity types, but as long as I can get a string entity name, that's all that matters.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AbstractDTO.java
jacksonargo: Ya, I wasn't really sure where to put this method. I think it makes sense to have a generic method like this and have resource/entity specific methods override it. Or maybe do a switch statement on the dto/api class.

jacksonargo: The conversions between dto -> api -> resource isn't super clear to me.

jacksonargo:
I'll borrow heavily from the pinot plugin for the implementation of the auth platform client https://github.com/startreedata/startree-pinot/tree/master/startree-plugins/startree-auth-platform/src/main/java/ai/startree/pinot/auth/platform

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/authorization/EntityType.java
jacksonargo: Is this what you had in mind? https://github.com/startreedata/thirdeye/pull/803/files#diff-eda7b7b48af5e20539ea96eab104586de9e4fc73adff1b44d0026b6d975338c6R20

jacksonargo:
Here's a rough outline for the auth platform implementation: https://github.com/startreedata/startree-thirdeye/pull/83

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/AccessControl.java
```diff
@@ -0,0 +1,26 @@
...
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.authorization;
+
+import ai.startree.thirdeye.auth.ThirdEyePrincipal;
+import java.net.http.HttpHeaders;
+
+public interface AccessControl {
```
suvodeep-pyne, line 19: regarding packaging. I think now we have `ai.startree.thirdeye.auth` and `ai.startree.thirdeye.authorization`.

Not sure keeping it flat is the best idea.
- either we can have all authorization classes in auth as well. so auth = authentication + authorization
- else, if you foresee more classes coming in for authorization, then we can further the hierarchy like `ai.startree.thirdeye.auth.authorization`

suvodeep-pyne (commented):
Great stuff @jacksonargo! this is awesome. 
I have a few comments. lmk if you need me. can chat up

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
```diff
@@ -66,6 +72,8 @@
 
   private static final Logger log = LoggerFactory.getLogger(CrudResource.class);
 
+  public AccessControl accessControl = AccessControlBuilder.build();
```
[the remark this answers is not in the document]
jacksonargo: We def _dont_ need a different instance for each resource class or entity. We should only need 1 instance. I don't know how to do the dependency injection tho 😅

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/AccessControl.java
jacksonargo: Ya, I wasn't too sure on this myself. I don't expect to need more authorization classes, so I'll put these in the auth pkg.

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/ResourceIdentifier.java
```diff
@@ -0,0 +1,53 @@
...
+  public ResourceIdentifier(String name, String namespace, String entityType) {
+    this.name = name;
+    this.namespace = namespace;
+    this.entityType = entityType;
+  }
+
+  static public ResourceIdentifier fromDto(AbstractDTO dto) {
+    if (dto instanceof AlertDTO) {
+      return new ResourceIdentifier(
+          ((AlertDTO) dto).getName(),
```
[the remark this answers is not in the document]
jacksonargo: I wanted to use the name so people can make autorization policies that reference the resource names. We also need to specify the entity type, altho this might be able to come from the dto's themselves.

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/ResourceIdentifier.java
```diff
@@ -0,0 +1,53 @@
...
+import ai.startree.thirdeye.spi.datalayer.dto.AbstractDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
+
+public class ResourceIdentifier {
+
+  static public final String DefaultNamespace = "default";
+
+  public final String name;
+  public final String namespace;
+  public final String entityType;
```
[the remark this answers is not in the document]
jacksonargo: The entity type is like `"alert"`, `"anomaly"`, etc. Would there be an easy way to get it that info directly from the dto?

suvodeep-pyne (approved):
LGTM

The resource identification by "name" looks a little bit tricky. I was thinking more on the lines of "id"

Other than that. this is great work!

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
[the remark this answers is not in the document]
suvodeep-pyne: seems like you nailed dependency injection 🎯

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/ResourceIdentifier.java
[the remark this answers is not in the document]
suvodeep-pyne: there is a map inside the persistence layer which has names. `SubEntityType` is the class I think. but we have to move things around a bit if we wanna use that.

thirdeye-server/src/main/java/ai/startree/thirdeye/authorization/ResourceIdentifier.java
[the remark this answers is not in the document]
suvodeep-pyne: Hmm. 

That is interesting and a valid point. However, the challenge here would be if somebody changes the name in thirdeye, things would break. not sure if there is an easy way to address that. at least with id, we know ids won't change.

cyrilou242:
great work!

PR #806, opened 2022-11-29
[server] expose endpoint to list all the available datasets to onboard

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1093

#### Description

New endpoint on `DataSourceResource` to get a list of all available datasets (tables in pinot) which can be onboarded. This info is needed in UI so that user can select the datasets to onboard from the list of available datasets.

endpoint : `/api/data-sources/{datasource name}/available-datasets`
response:
```
[
  "Table_1",
  "Table_2",
  ...
  "Table_n"
]
```

### DIFF
--- a/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotThirdEyeDataSource.java
+++ b/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotThirdEyeDataSource.java
@@ -169,7 +169,11 @@ public class PinotThirdEyeDataSource implements ThirdEyeDataSource {
 
   @Override
   public List<String> getDatasets() throws Exception {
-    throw new UnsupportedOperationException();
+    try {
+      return datasetOnboarder.getAllTables();
+    } catch (final IOException e) {
+      throw new RuntimeException(e);
+    }
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotThirdEyeDataSource.java
```diff
@@ -169,7 +169,11 @@ public ThirdEyeResultSetGroup executeSQL(final PinotQuery pinotQuery) throws Exe
 
   @Override
   public List<String> getDatasets() throws Exception {
-    throw new UnsupportedOperationException();
```
jacksonargo, line 172: Should we remove `throws Exception` from the function declaration?

suvodeep-pyne:
@shounakmk219 
i want this done differently actually. the `onboardAllTables` actually only does get right now. so, that needs to be refactored. Is it Ok if I assign this back to myself. there is some context here which is missing on your end.

Also, I'm not sure if just returning the names of tables is a good idea. it is fine for now. However, in the long run, might be best to expose the dataset details in the same API.

shounakmk219:
@suvodeep-pyne Sure, fine by me.

suvodeep-pyne:
@shounakmk219 
This API is now available and merged. we can skip this PR

PR #823, opened 2022-12-05
[detectionpipeline] remove duplicated merging logic

cyrilou242:
## Major changes:

- introduce DetectionPipelineContext$usage
usage is in `DETECTION`, `EVALUATION`
--> similar to `preserveOutputDataFrames` but less specific, let the implementation decide what should be done. (Maybe could replace `preserveOutputDataFrames`?).
for AnomalyMerger: 
  - if `usage=DETECTION`, existing anomalies in persistence are pulled and merged with the pipeline anomalies. Used in detectionTaskRunner flow
  - if `usage=EVALUATION`, existing anomalies are not pulled. Used in /evaluate flow

- simplify merging logic for dimension exploration
  merging happens in each forked process
  enumeration item is obtained from the DetectionPipelineContext. 

- alertId is passed in the detectionPipelineContext
  allows fetching anomalies of the alert in the pipeline.
  
  
  
## Review pointers: 
- Look at how [DetectionPipelineUsage](https://github.com/startreedata/thirdeye/pull/823/files#diff-cac70bcc1f51c3e7d960b34728b2c5e90b5c38d0079d7e2e098ab2e5c29f9953R81) is set and used
- Look at AnomalyMergerPostProcessor
- Look at one [template](https://github.com/startreedata/thirdeye/pull/823/files#diff-d9a34e2608665786dcd0744b3a55fcad2890433d50d2c931c1e1d9a69a05c451R8)

## Risk
- will break merge behavior of custom templates (startree bits clients)


thirdeye-server/src/main/java/ai/startree/thirdeye/alert/AlertEvaluator.java
```diff
@@ -88,6 +89,8 @@ public AlertEvaluationApi evaluate(final AlertEvaluationApi request)
           detectionInterval);
 
       final DetectionPipelineContext context = new DetectionPipelineContext()
+          .setAlertId(request.getAlert().getId())
+          .setUsage(DetectionPipelineUsage.EVALUATION)
```
suvodeep-pyne, line 93: Just FYI. I think we are creating different names for the same thing. 

In `DetectionPipelineContext`
```  
private boolean preserveOutputDataFrames = false;
````

this variable is used to understand whether we need to preserve the big data frames.

Comment (cyrilou242) [reply thirdeye-server/src/main/java/ai/startree/thirdeye/alert/AlertEvaluator.java:93]:
```
private boolean preserveOutputDataFrames = false;
```

yes saw this one. 
I could have used this boolean but the name was specific to the forkjoin logic, and I'm more confident with an enum, in the future we could allow things like:
EVALUATE_AND_DROP_DATAFRAMES (for some perf opti?)
or 
DETECT_KEEP_DATAFRAME (to store historical data in persistence/generate preview in notifications?)
DRIFT_DETECTION?
etc...

I'll drop `preserveOutputDataFrames` in favor of `usage` in another PR if it's okay for you.

PR #829, opened 2022-12-09
[kubernetes] sample dataset ingestion

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1094

#### Description

To enable sample data creation along with the ThirdEye installation we need to provide a helper job that uploads the sample datasets to the user provided pinot instance so that this data can be used later to create sample alerts for first time users.

This PR introduces :

1. A k8s `Job` that checks pinot connectivity and then pushes sample data to it.
2. A k8s `ConfigMap` that holds the script which the job runs to upload data.

values.yaml properties : 

- `dataloader.enabled` : flag to enable/disable the sample data ingestion job. Default - `false` 
- `dataloader.pinot.controllerProtocol` : pinot controller protocol http/https. Default - `http`
- `dataloader.pinot.controllerHost` : pinot controller host dns. Default - `pinot-pinot-controller-headless`
- `dataloader.pinot.controllerPort` : pinot controller port. Default - `9000`
- `dataloader.pinot.commandOptions` : any extra command line options to pass to the pinot-admin.sh tool such as if auth is enabled on pinot then pass `-authToken 'Basic <auth token>'`. Default - none
- `dataloader.sampleDatasets` : it expects an array of sample dataset base urls. Please follow the sample dataset directory structure as followed by the existing datasets in examples directory of this repo. The base url should be such that files must be downloadable over http, for example `https://raw.githubusercontent.com/startreedata/thirdeye/master/examples/order_events`

#### How to test

1. Run a pinot instance in local
2. Pass these dataloader values to the values.yaml file
```
dataloader:
  enabled: true
  pinot:
    controllerHost: <machine ip such as 192.168.1.2>
  sampleDatasets:
    - https://raw.githubusercontent.com/startreedata/thirdeye/master/examples/order_events
    - https://raw.githubusercontent.com/startreedata/thirdeye/master/examples/pageviews
```
3. Install the helm charts
4. Once the pod for the dataloader job goes to completion check pinot for the tables


cyrilou242:
tested on my machine, works fine 👍 
used a local pinot on the minikube host and used 
```
controllerHost:host.minikube.internal 
```

solution wise I understand you already discussed it with Suvodeep, LGTM.

kubernetes/helm/startree-thirdeye/templates/quickstart/job.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: wait-for-pinot
+        image: alpine/curl:latest
+        env:
+          - name: CONTROLLER_PROTOCOL
+            value: {{ (.Values.quickstart.pinot).controllerProtocol | default "http" | quote }}
+          - name: CONTROLLER_HOST
+            value: {{ (.Values.quickstart.pinot).controllerHost | default "pinot-pinot-controller-headless" | quote }}
+          - name: CONTROLLER_PORT
+            value: {{ (.Values.quickstart.pinot).controllerPort | default "9000" | quote }}
+        command: ['sh', '-c', 'until curl -f $CONTROLLER_PROTOCOL://$CONTROLLER_HOST:$CONTROLLER_PORT/health; do echo waiting for pinot; sleep 2; done']
```
cyrilou242, line 36: I think a timeout is necessary. 
At first my configuration was incorrect and the pod got blocked in init state forever, without much feedback
I think a 5 minutes timeout then an error that prints the pinot configuration and that says the pinot instance could not be reached would help.

shounakmk219: Valid point. Was worried about the job exceeding the retry limits before the issue being resolved on pinot, then we need to manually trigger it. But if the configs that are passed are in itself incorrect then anyways we need to manually delete the old job first before pushing the upgrade as Job resource is read only in k8s. But we should be able to hack our way through both the cases by setting `quickstart.enabled=false` and then set it back to `quickstart.enabled=true` along with updates to pinot configs if any. 
Thanks for pointing this out!

shounakmk219: Added the timeout with descriptive log.

PR #840, opened 2022-12-15
fix(ui): TE-1147 fix bug where UI should send non quoted strings

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1147

#### Description

Adding parameter for `concatKeyValueWithEqual` to quote strings or not so it can be used by the new feature that introduced the bug (cohort recommender) and the old code

#### Screenshots

This is more of a code change than a UI change


alcatraz627 (approved):

alcatraz627:
lgtm

PR #845, opened 2022-12-19
Add access control for remaining alert actions: create, update, validate, etc.

jacksonargo:
This pr adds access control hooks for POST/PUT and the alert specific actions.


thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ResourceIdentifier.java
```diff
@@ -15,37 +15,50 @@
...
   public final String namespace;
-  public final String entityType;
+  public final EntityType entityType;
 
-  public ResourceIdentifier(final String name, final String namespace, final String entityType) {
+  public ResourceIdentifier(final String name, @Nullable final String namespace,
+      final EntityType entityType) {
     this.name = name;
-    this.namespace = namespace;
+    this.namespace = namespace != null ? namespace : DefaultNamespace;
```
suvodeep-pyne, line 32: nit: use optional()
```
optional(namespace).orElse(DEFAULT_NAMESPACE)
```

why? it looks cute :)

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ResourceIdentifier.java
```diff
@@ -15,37 +15,50 @@
 
 import ai.startree.thirdeye.spi.datalayer.dto.AbstractDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AlertTemplateDTO;
+import javax.annotation.Nullable;
 
 public class ResourceIdentifier {
 
   static public final String DefaultNamespace = "default";
```
suvodeep-pyne, line 23: we mostly follow Google coding conventions.

constants should be capitalized and underscored. see Constants.java
```
  static public final String DEFAULT_NAMESPACE = "default";
```

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ResourceIdentifier.java
```diff
@@ -15,37 +15,50 @@
...
 
   public final String name;
   public final String namespace;
-  public final String entityType;
+  public final EntityType entityType;
 
-  public ResourceIdentifier(final String name, final String namespace, final String entityType) {
+  public ResourceIdentifier(final String name, @Nullable final String namespace,
```
suvodeep-pyne, line 29: `@Nullable` is nice. but how are we in general planning to use the namespace attribute?

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ResourceIdentifier.java
```diff
@@ -15,37 +15,50 @@
...
+      final EntityType entityType) {
     this.name = name;
-    this.namespace = namespace;
+    this.namespace = namespace != null ? namespace : DefaultNamespace;
     this.entityType = entityType;
   }
 
   static public ResourceIdentifier fromDto(final AbstractDTO dto) {
```
suvodeep-pyne, line 36: Question here: do we need multiple public interfaces here? or a single `resourceId(entity)` should suffice

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
```diff
@@ -213,6 +230,8 @@ public Response validateMultiple(
       final AlertDTO existing =
           api.getId() == null ? null : ensureExists(dtoManager.findById(api.getId()));
       validate(api, existing);
+      ensureHasAccess(principal, api, AccessType.UPDATE);
+      ensureHasAccessToRelatedResources(principal, api);
```
suvodeep-pyne, line 234: Not sure I understand this? So user should have access to alert template when getting alert?

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
```diff
@@ -110,6 +113,11 @@ private DtoT updateDto(final ThirdEyePrincipal principal, final ApiT api) {
 
     final DtoT updated = toDto(api);
 
+    // Check the that user can modify the resource before and after the update.
```
suvodeep-pyne, line 116: Longer term we need to refactor this setup entirely I feel. but thanks for writing this in such a concise fashion. 👍

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
```diff
@@ -206,11 +214,36 @@ public boolean hasAccess(final ThirdEyePrincipal principal, final AbstractDTO dt
 
   public void ensureHasAccess(final ThirdEyePrincipal principal, final AbstractDTO dto,
```
suvodeep-pyne, line 215: Is there a way we can move all of this logic outside? meaning we create a generic service class which has all these methods and instances to decide it all and let `CrudResource` know? This is a lot of dedicated code for auth which could be containerized? thoughts?

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
```diff
@@ -206,11 +214,36 @@ public boolean hasAccess(final ThirdEyePrincipal principal, final AbstractDTO dt
...
 
+  public final void ensureHasAccessToRelatedResources(ThirdEyePrincipal principal, ApiT api) {
+    ensureHasAccessToRelatedResources(principal, toDto(api));
+  }
+
+  public final void ensureHasAccessToRelatedResources(ThirdEyePrincipal principal, DtoT dto) {
+    relatedDtos(dto).forEach(related -> ensureHasAccess(principal, related, AccessType.READ));
+  }
+
+  public List<ResourceIdentifier> relatedDtos(final DtoT dto) {
```
suvodeep-pyne, line 243: nice! 

but this seems like an auth interface. need to see how to design this. The minus I see here is that Resource classes have to say how auth behaves in entities. don't think thats the best place for it.

Resource classes should ideally just collect data from services or push requests into classes which hold more of the business logic.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AlertTemplateDTO.java
```diff
@@ -23,6 +23,7 @@
 public class AlertTemplateDTO extends AbstractDTO {
 
   private String name;
+  private String namespace;
```
suvodeep-pyne, line 26: This could be tricky. 

Either this should be at the top level like `AbstractDTO` since, potentially most objects will have some kind of auth.

Another question: Is `namespace` the only identifier and entity needed for auth in the foreseeable future? Else would suggest creating a container here instead of having it as `String`.

suvodeep-pyne (commented):
Great stuff again @jacksonargo some comments and few questions here.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ResourceIdentifier.java
jacksonargo: Omg I looked for something like this haha! Ya much nicer.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AlertTemplateDTO.java
jacksonargo: We only need id/name, namespace, and entity type, and that's all for the foreseeable future. Namespace on AbstractDTO makes sense to me.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/AlertResource.java
jacksonargo: The idea is that a user can't create an alert if they don't have access to the alert template. Similarly, for dataset+datasource. But I'm not sure if that makes sense or if this is the best way to implement it.

I think I'll hold off on the related resources business for now.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/CrudResource.java
jacksonargo: 🤔 Ya, I agree it needs it's own class. I'm starting to think a wrapper on top of the data managers might be the way to go. For example, when CrudResource calls dataManager.findById(), it's an auth wrapper which calls realManager.findByID() and does the access control check.

This would isolate all the auth code from CrudResource and also probably save me from missing an endpoint.

jacksonargo: Decided not to go w/ wrapper ... was way to clunky haha. I made a class AuthorizationManager to handle the this code.

suvodeep-pyne (approved):
LGTM

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -0,0 +1,113 @@
...
+
+  public boolean hasAccess(final ThirdEyePrincipal principal,
+      final ResourceIdentifier identifier, final AccessType accessType) {
+    return accessControl.hasAccess(principal, identifier, accessType);
+  }
+
+  private <T extends AbstractDTO> List<ResourceIdentifier> relatedEntities(T entity) {
+    if (entity instanceof AlertDTO) {
+      final AlertDTO alertDto = (AlertDTO) entity;
+      final AlertTemplateDTO alertTemplateDTO = new AlertTemplateRenderer(
```
cyrilou242, line 95: what is the reason for creating a templateRenderer instead of using the singleton?

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -0,0 +1,113 @@
...
+          alertManager, alertTemplateManager).getTemplate(alertDto.getTemplate());
```
cyrilou242, line 96: `alertDto.getTemplate()` will return null when only the id of the alert is passed. 
(eg in the evaluate call in anomalies page https://thirdeye.demo.teprod.startree.cloud/anomalies/6978/view?timeRange=CUSTOM&startTime=1619654400000&endTime=1622160000000)
this breaks the anomalies page, throwing NPE. 
this was not caught by integration tests :( --> I'll add a test for this.

without refactoring getTemplate should stay private and only renderAlert(alertDto).getTemplate() is a safe way to get the template. 
we can think about refactoring this if the rendering takes too much time.

cyrilou242 (commented):
late in the game but awesome work! 
put some comments here because I think some issues are caused by this PR. 
I'll create a PR and tag you @jacksonargo.

PR #855, opened 2022-12-28
chore(ui): TE-1168 update template properties default values

cyrilou242:
timeColumn and timeColumnFormat are inferred automatically now, no need to set them.


alcatraz627 (approved):
lgtm

PR #860, opened 2023-01-03
te 1172 fix access control npe

cyrilou242:
Note: 

- in https://github.com/startreedata/thirdeye/pull/860/files#diff-bb8afb0991edd6ea1a1df622969ccbfec1f23bcb88882f88c888416a144e8af1 the transformation from incomplete DTO to complete DTO is made by the resource
- in https://github.com/startreedata/thirdeye/pull/860/files#diff-9fd90e5ff69e9a40acedb4864d6baa0d480df290cc18ca613f7866a5446bf6a7R89 the transformation from incomplete DTO to complete DTO is made by the authorizationManager 

--> it's not clear yet to me which level should be responsible for this @jacksonargo. I think this will be an important design consideration to avoid bugs.

### DIFF
--- a/thirdeye-integration-tests/src/test/java/ai/startree/thirdeye/HappyPathTest.java
+++ b/thirdeye-integration-tests/src/test/java/ai/startree/thirdeye/HappyPathTest.java
@@ -246,6 +246,17 @@ public class HappyPathTest {
   }
 
   @Test(dependsOnMethods = "testGetAnomalies")
+  public void testEvaluateWithAlertIdOnly() {
+    // corresponds to the evaluate call performed from an anomaly page - only the alertId is passed
+    final AlertEvaluationApi alertEvaluationApi = alertEvaluationApi(new AlertApi().setId(alertId),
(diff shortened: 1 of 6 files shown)


cyrilou242:
will move the alertTemplate fetching down to the authorizationManager

jacksonargo:
Alert resource seems to be the best place to fetch the complete dto, and the complete dto is passed to the authorization manager and template renderer.

jacksonargo:
The AlertTemplateRenderer.renderAlert() method  seems to get the alert template from the alert api object, which is a little confusing if the alert api objects sometimes only have the id.

https://github.com/startreedata/thirdeye/blob/f813c2611a90293945d0a2753e520941317c7bcf/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertTemplateRenderer.java#L71-L72

cyrilou242:
ok let me remove my two last commits

jacksonargo (approved):
:shipit:

PR #861, opened 2023-01-03
relatedEntities() returns empty list if alert template is null

jacksonargo:
Fixes issue in https://github.com/startreedata/thirdeye/pull/845#discussion_r1060453161


cyrilou242:
Hey @jacksonargo I don't think this solves the issue.
The NPE happens inside the getTemplate call.

The input of evalute can be: 

```
{id: 1234}
```

In this case, I think the authorization manager should: 
- fetch the complete alert object
- return the template for access control

Could you have a look at: 
https://github.com/startreedata/thirdeye/pull/860

PR #872, opened 2023-01-09
Add plugin method to set up access control

jacksonargo:
This pr add a method providing an AccessControl instance to the Plugin interface. 

There's a generic access control config that has a flag for enabled/disabled and a map of plugin specific properties. 

I moved the AccessControl interface and related resources into the spi package so Plugin can reference them.

A single AccessControlProvider class is bound to the AccessControl interface. The Pugin loader can store a new AccessControl instance in the AccessControlProvider, which is subsequently used by server resources. 

I verified that swapping in different access control instances from the plugin loader does work as expected. I also tested w/ https://github.com/startreedata/startree-thirdeye/pull/93 to make sure that the plugin loads and runs correctly.

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
@@ -16,6 +16,7 @@ package ai.startree.thirdeye;
 import static ai.startree.thirdeye.spi.util.SpiUtils.optional;
 import static java.util.Objects.requireNonNull;
 
+import ai.startree.thirdeye.auth.AccessControlProvider;
 import ai.startree.thirdeye.core.BootstrapResourcesRegistry;
 import ai.startree.thirdeye.datasource.DataSourcesLoader;
 import ai.startree.thirdeye.detectionpipeline.DetectionRegistry;
@@ -65,6 +66,7 @@ public class PluginLoader {
   private final ContributorsFinderRunner contributorsFinderRunner;
   private final BootstrapResourcesRegistry bootstrapResourcesRegistry;
   private final PostProcessorRegistry postProcessorRegistry;
+  private final AccessControlProvider accessControlProvider;
 
   private final AtomicBoolean loading = new AtomicBoolean();
   private final File pluginsDir;
@@ -77,13 +79,15 @@ public class PluginLoader {
       final ContributorsFinderRunner contributorsFinderRunner,
       final BootstrapResourcesRegistry bootstrapResourcesRegistry,
       final PostProcessorRegistry postProcessorRegistry,
-      final PluginLoaderConfiguration config) {
+      final PluginLoaderConfiguration config,
+      final AccessControlProvider accessControlProvider) {
     this.dataSourcesLoader = dataSourcesLoader;
     this.detectionRegistry = detectionRegistry;
     this.notificationServiceRegistry = notificationServiceRegistry;
     this.contributorsFinderRunner = contributorsFinderRunner;
     this.bootstrapResourcesRegistry = bootstrapResourcesRegistry;
     this.postProcessorRegistry = postProcessorRegistry;
+    this.accessControlProvider = accessControlProvider;
     pluginsDir = new File(config.getPluginsPath());
   }
 
@@ -143,6 +147,7 @@ public class PluginLoader {
     for (EnumeratorFactory f : plugin.getEnumeratorFactories()) {
       detectionRegistry.addEnumeratorFactory(f);
     }
+    optional(plugin.getAccessControl()).ifPresent(accessControlProvider::set);
 
     log.info("Installed plugin: " + plugin.getClass().getName());
   }
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
@@ -79,8 +79,8 @@ public class PluginLoader {
       final ContributorsFinderRunner contributorsFinderRunner,
       final BootstrapResourcesRegistry bootstrapResourcesRegistry,
       final PostProcessorRegistry postProcessorRegistry,
-      final PluginLoaderConfiguration config,
-      final AccessControlProvider accessControlProvider) {
+      final AccessControlProvider accessControlProvider,
+      final PluginLoaderConfiguration config) {
     this.dataSourcesLoader = dataSourcesLoader;
     this.detectionRegistry = detectionRegistry;
     this.notificationServiceRegistry = notificationServiceRegistry;
@@ -147,8 +147,10 @@ public class PluginLoader {
     for (EnumeratorFactory f : plugin.getEnumeratorFactories()) {
       detectionRegistry.addEnumeratorFactory(f);
     }
-    optional(plugin.getAccessControl(accessControlProvider.getConfig()))
-        .ifPresent(accessControlProvider::setAccessControl);
+    if (accessControlProvider.getConfig().enabled) {
+      optional(plugin.getAccessControl(accessControlProvider.getConfig()))
+          .ifPresent(accessControlProvider::setAccessControl);
+    }
 
     log.info("Installed plugin: " + plugin.getClass().getName());
   }
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
@@ -35,12 +35,14 @@ import ai.startree.thirdeye.spi.notification.NotificationServiceFactory;
 import ai.startree.thirdeye.spi.rca.ContributorsFinderFactory;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import com.google.inject.name.Named;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URI;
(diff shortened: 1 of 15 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
```diff
@@ -77,14 +82,20 @@ public PluginLoader(
...
     this.detectionRegistry = detectionRegistry;
     this.notificationServiceRegistry = notificationServiceRegistry;
     this.contributorsFinderRunner = contributorsFinderRunner;
     this.bootstrapResourcesRegistry = bootstrapResourcesRegistry;
     this.postProcessorRegistry = postProcessorRegistry;
+    this.accessControlProvider = accessControlProvider;
+    this.accessControlConfiguration = accessControlConfiguration;
     pluginsDir = new File(config.getPluginsPath());
+
+    System.out.println(accessControlConfiguration);
```
jacksonargo, line 98: lmaoooooooo

thirdeye-server/src/main/java/ai/startree/thirdeye/ThirdEyeServerModule.java
```diff
@@ -71,11 +72,14 @@ protected void configure() {
     install(new ThirdEyeDetectionPipelineModule(configuration.getDetectionPipelineConfiguration()));
     install(new ThirdEyeWorkerModule(configuration.getTaskDriverConfiguration()));
     install(new ThirdEyeSchedulerModule(configuration.getSchedulerConfiguration()));
-    install(new AccessControlModule());
-
     bind(AuthConfiguration.class).toInstance(configuration.getAuthConfiguration());
     bind(MetricRegistry.class).toInstance(metricRegistry);
     bind(ThirdEyeServerConfiguration.class).toInstance(configuration);
+
+    final AccessControlProvider accessControlProvider = new AccessControlProvider(
```
suvodeep-pyne, line 79: Instead. just use a provider method. see examples in the methods below

thirdeye-server/src/main/java/ai/startree/thirdeye/PluginLoader.java
```diff
@@ -77,13 +79,15 @@ public PluginLoader(
       final ContributorsFinderRunner contributorsFinderRunner,
       final BootstrapResourcesRegistry bootstrapResourcesRegistry,
       final PostProcessorRegistry postProcessorRegistry,
-      final PluginLoaderConfiguration config) {
+      final PluginLoaderConfiguration config,
```
suvodeep-pyne, line 82: nit: let's make `PluginLoaderConfiguration`, the first or the last dependency.

suvodeep-pyne: fyi. This is purely cosmetic. was thinking it would be more organized to have the config and then all the registries or vice versa. However, having the config somewhere in the middle would be a bit iffy

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AccessControlProvider.java
```diff
@@ -0,0 +1,73 @@
...
+    this.accessControl = accessControl;
+  }
+
+  public AccessControl getAccessControl() {
+    if (!config.enabled) {
+      return alwaysAllow;
+    }
+
+    if (this.accessControl == null) {
+      throw new RuntimeException("Access control is enabled, but no provider has been configured!");
```
suvodeep-pyne, line 59: perfect. 

nit: you can also use `checkState()`, checkState is a static method from Preconditions. there are sample usages in our codebase for this

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AccessControlProvider.java
```diff
@@ -0,0 +1,73 @@
...
+  ) -> true;
+
+  public final static AccessControl alwaysDeny = (
+      final String token,
+      final ResourceIdentifier identifiers,
+      final AccessType accessType
+  ) -> false;
+
+  private final AccessControlConfiguration config;
+  private AccessControl accessControl = null;
```
suvodeep-pyne, line 40: nit: (For Later) `AccessControl` object is actually the class that checks/enforces the policy. The name is somehow not conveying this. don't have a great suggestion for this. I was thinking `AccessControlEnforcer` or `AccessControlPolicyChecker` or whatever :|

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -72,23 +82,41 @@ public void ensureCanValidate(final ThirdEyePrincipal principal, final AlertDTO
...
+    return hasAccess(principal, dtoToId(entity), accessType);
+  }
+
   public boolean hasAccess(final ThirdEyePrincipal principal,
       final ResourceIdentifier identifier, final AccessType accessType) {
-    return accessControl.hasAccess(principal, identifier, accessType);
+    return accessControl.hasAccess(principal.authToken, identifier, accessType);
+  }
+
+  static public <T extends AbstractDTO> ResourceIdentifier dtoToId(final T dto) {
```
suvodeep-pyne, line 107: nice! I'd just call it `resourceIdentifier(entity)` or `resourceId(entity)`. `dtoToId` is a bit confusing since `id` is also the database id and doesn't call out resourceId

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/Plugin.java
```diff
@@ -56,4 +59,8 @@ default Iterable<AnomalyPostProcessorFactory> getAnomalyPostProcessorFactories()
   default Iterable<EnumeratorFactory> getEnumeratorFactories() {
     return Collections.emptyList();
   }
+
+  default AccessControl getAccessControl(AccessControlConfiguration config) {
```
suvodeep-pyne, line 63: nit: default should be just `AlwaysAllow`? or `null` and `AlwaysAllow` mean the same? just trying to simplify.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/accessControl/ResourceIdentifier.java
```diff
@@ -29,18 +25,18 @@ public class ResourceIdentifier {
...
 
-  static public <T extends AbstractDTO> ResourceIdentifier from(final T dto) {
+  public static ResourceIdentifier from(final String name, final String namespace,
+      final String entityType) {
     return new ResourceIdentifier(
-        optional(dto.getId()).map(Objects::toString).orElse(DEFAULT_NAME),
-        optional(dto.getNamespace()).orElse(DEFAULT_NAMESPACE),
-        optional(SubEntities.BEAN_TYPE_MAP.get(dto.getClass()))
-            .map(Objects::toString).orElse(DEFAULT_ENTITY_TYPE)
+        optional(name).orElse("").length() > 0 ? name : DEFAULT_NAME,
```
suvodeep-pyne, line 37: nit: you can use `filter()` if you'd like

so something like 
```
optional(name)
        .filter(StringUtils::isNotEmpty)
        .orElse(DEFAULT_NAME);
```

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/accessControl/AccessControlConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package ai.startree.thirdeye.spi.accessControl;
+
+import java.util.Map;
+
+public class AccessControlConfiguration {
+  public boolean enabled;
```
suvodeep-pyne, line 20: We may need to think a bit more about the config structure. this is a plugin config that we are trying to pass on here. am i right?

This is a rough take. but I think containerizing this might be beneficial in the long run.
```
# inside server.yaml

auth: # Auth Config
  enabled: true
  # other auth conf

  plugins: # AuthPluginConfigs <String, Map<>>
    startree-auth: # only this map gets pushed to the plugin.
      key1: value1
      key2: value2

```

suvodeep-pyne (commented):
LGTM. mostly nit picks in comments that mostly doesn't matter except 1. config design.. please take a look and lmk what you think

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/accessControl/AccessControlConfiguration.java
jacksonargo: Ya, this format looks good. (tbh I was being a little lazy w/ design 😅 ) Instead of passing only the `startee-auth` field, I'll pass the whole `plugins` field and the plugin itself looks for the `startree-auth` field. This way the plugin loader doesn't have to know how to map config entries to individual plugins.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/Plugin.java
jacksonargo: This is the default method for all plugins, so null here indicates that the plugin does not provide access control. The other default methods return empty lists, but I don't think list makes sense for an access control plugin, so I chose null as the nothing here value.

suvodeep-pyne (approved):
LGTM.

PR #873, opened 2023-01-09
Add rideshare demo data

jacksonargo:
This pr adds the table and schema for rideshare demo data.

ref https://cortexdata.atlassian.net/browse/TE-1174


cyrilou242:
could you add a small README in examples/rideshare that 
- explains the dataset
- says if it's for testing or use case demo
- mention the source and the license if there is one? I wonder where Madhu got this.

we haven't been consistent at documenting datasets and now we have a bunch of datasets nobody knows what they are good at. 

also have you tried an alert on it? Is it an interesting dataset?

PR #882, opened 2023-01-16
feat(ui): TE-1211 deleting timestamp keys from copied alert

alcatraz627:
#### Issue(s)

[TE-1211](https://cortexdata.atlassian.net/browse/TE-1211)

#### Description

- Removed `created` and `updated` key from duplicated alerts when in editable mode

#### Screenshots

Screenshot_45

[TE-1211]: https://cortexdata.atlassian.net/browse/TE-1211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thirdeye-ui/src/app/pages/alerts-create-page/alerts-create-copy-page.component.tsx
```diff
@@ -29,11 +29,20 @@ export const AlertsCreateCopyPage: FunctionComponent = () => {
...
         getAlert(Number(alertId)).then((alert) => {
             if (alert) {
+                const keysToDelete: (keyof Alert)[] = [
+                    "id",
+                    "lastTimestamp",
+                    "active",
+                    "owner",
+                    "created",
+                    "updated",
+                ];
```
spham92, line 39: Probably best to make this a constant before the declaration of the component at line 22 so we don't have to reinitialize this every time

alcatraz627: I moved the whole copying logic (the deletion code and these keys) to an accompanying utils file to hide the implementation detail, and to allow for future reuse if needed.

thirdeye-ui/src/app/pages/alerts-create-page/alerts-create-page.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    "updated",
+];
+
+export const createAlertCopy = (alert: Alert): EditableAlert => {
+    const editableVersion: EditableAlert = { ...alert };
+    keysToDelete.forEach((k) => {
+        delete (editableVersion as Alert)[k];
+    });
+
+    return editableVersion;
```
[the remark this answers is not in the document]
alcatraz627: Added a test case.

In addition to normal alert copying, also added a case where an editable alert can be copied as well, if passed as a parameter

PR #891, opened 2023-01-19
fix(ui): TE-1228 use existing value for sensitivity if it exists for the slider in new user flow

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1228

#### Description

If sensitivity value exists in the alert template properties, use it as the default value

#### Screenshots

No UI changes


alcatraz627 (approved):
lgtm

PR #896, opened 2023-01-20
update mappers to support namespace field

jacksonargo:
In previous prs, I added the namespace field to the AbstractDTO class and the AlertAPI and AlertTemplateAPI classes. I didn't update the mappers, so the namespace never made it to the database 🤦

Update api<->dto mappers to support namespace field in Alerts, Datasets, Data sources.

Tested by adding namespace field to the resources.


cyrilou242:
:shipit:

PR #898, opened 2023-01-23
Set namespaces for anomalies

jacksonargo:
Anomalies inherit the authorization config from its parent resource.

- If the Anomaly comes from an Alert, then it gets a copy of the Alerts auth config.
- If the Anomaly comes from an Enumeration Item, and the Enumeration Item has a non-null auth config, then the Anomaly gets a copy of the Enumeration Item's auth config. Otherwise, the Anomaly gets a copy of the Alert's auth config.
  - ~If the Alert has an non-null auth config, but the Enumeration item has null auth config, then the Anomaly will still have a null auth config. I considered using the Alert's non-null auth config in this case, but I felt that it adds an extra layer of complexity with little value.~

This change will affect new anomalies, but old anomalies will still have a null auth config. I make a tool to update the auth configs for existing anomalies in a separate pr.

Tested by setting a namespace on alerts and inspecting the namespace for the resulting anomalies.

https://cortexdata.atlassian.net/browse/TE-1255

### DIFF
--- a/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
+++ b/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
@@ -27,6 +27,7 @@ public class DetectionPipelineContext {
   private EnumerationItemDTO enumerationItem;
   private DetectionPipelineUsage usage;
   private Long alertId;
+  private String namespace;
 
   public ApplicationContext getApplicationContext() {
     return applicationContext;
@@ -82,6 +83,15 @@ public class DetectionPipelineContext {
 
   public DetectionPipelineContext setAlertId(final Long alertId) {
     this.alertId = alertId;
+    return this;
+  }
+
+  public String getNamespace() {
+    return namespace;
+  }
+
+  public DetectionPipelineContext setNamespace(final String namespace) {
+    this.namespace = namespace;
     return this;
   }
 }
--- a/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
+++ b/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
@@ -14,7 +14,6 @@
 package ai.startree.thirdeye.detectionpipeline;
 
 import ai.startree.thirdeye.spi.datalayer.Predicate;
-import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.EnumerationItemDTO;
 import ai.startree.thirdeye.spi.detection.DetectionPipelineUsage;
 import java.util.List;
@@ -28,7 +27,6 @@ public class DetectionPipelineContext {
   private EnumerationItemDTO enumerationItem;
   private DetectionPipelineUsage usage;
   private Long alertId;
-  private AuthorizationConfigurationDTO anomalyAuth;
 
   public ApplicationContext getApplicationContext() {
     return applicationContext;
@@ -84,15 +82,6 @@ public class DetectionPipelineContext {
 
   public DetectionPipelineContext setAlertId(final Long alertId) {
     this.alertId = alertId;
-    return this;
-  }
-
-  public AuthorizationConfigurationDTO getAnomalyAuth() {
-    return anomalyAuth;
-  }
-
-  public DetectionPipelineContext setAnomalyAuth(final AuthorizationConfigurationDTO auth) {
-    this.anomalyAuth = auth;
     return this;
   }
 }
--- a/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
+++ b/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
@@ -13,10 +13,8 @@
  */
 package ai.startree.thirdeye.detectionpipeline;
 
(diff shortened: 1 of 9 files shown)


jacksonargo:
It might be better to use the alert namespace all together. Not 100% sure either way https://github.com/startreedata/thirdeye/pull/909

suvodeep-pyne (commented):
Is there a reason that we need to do this in the detection pipeline? or should we do after the pipeline execution is complete. 

IMO slapping an auth tag shouldn't be related to the detection pipeline. thoughts?

suvodeep-pyne:
@cyrilou242 @jacksonargo thoughts?

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
```diff
@@ -27,6 +28,7 @@ public class DetectionPipelineContext {
   private EnumerationItemDTO enumerationItem;
   private DetectionPipelineUsage usage;
   private Long alertId;
+  private AuthorizationConfigurationDTO anomalyAuth;
```
suvodeep-pyne, line 31: Also we should have `alert` here if we are going to do it this way. 
What I don't like is the modularity of the detection pipeline is being broken if we are introducing db level operations in the detection pipeline.
@cyrilou242 @jacksonargo

jacksonargo:
> Is there a reason that we need to do this in the detection pipeline? or should we do after the pipeline execution is complete.

I took this approach since the detection pipeline is where the context object for the enum items is created. It seemed like good place to stash the auth config.

After pipeline execution makes sense to me. Maybe as a part of the post-processor operator or as a separate operator that just sets the auth config?

cyrilou242:
>If the Anomaly comes from an Enumeration Item, then it always gets a copy of the Enumeration Item's auth config.
If the Alert has an non-null auth config, but the Enumeration item has null auth config, then the Anomaly will still have a null auth config. I considered using the Alert's non-null auth config in this case, but I felt that it adds an extra layer of complexity with little value.

Not sure to understand. Having to set the auth for each enum is a pain if I want the same access control for each enum right?
Use case: I am a data eng, I have a dimension exploration of devices. I want to see all of them. 
But the business teams should not see this alert and the anomalies.
I just want to set the auth config at the alert level.

thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/DetectionPipelineContext.java
cyrilou242: agree, seems to be fixed in the newest commits

jacksonargo:
> Having to set the auth for each enum is a pain if I want the same access control for each enum right?

I'm really torn on this. The idea of inheriting config from multiple places feels hard to reason about, but intuitively, it seems like the expected outcome.

In my first implementation, the logic for copying the auth config was spread across the alert evaluator and enum item evaluator. Now that the auth config is set just before the anomalies are written to the db, the logic happens in one place, so it's much easier see where the auth config comes from.

jacksonargo:
I'll change the logic to use the Alert namespace if the enum item doesn't have one.

thirdeye-worker/src/main/java/ai/startree/thirdeye/worker/task/runner/DetectionPipelineTaskRunner.java
```diff
@@ -199,4 +206,19 @@ private void postExecution(final OperatorResult result) {
       }
     }
   }
+
+  private AnomalyDTO setAnomalyAuth(final AnomalyDTO anomaly) {
```
jacksonargo, line 210: Is there a better place for this method?

suvodeep-pyne: Let's start here and refactor as needed. for now lgtm

thirdeye-worker/src/main/java/ai/startree/thirdeye/worker/task/runner/DetectionPipelineTaskRunner.java
```diff
@@ -183,12 +190,12 @@ private void postExecution(final OperatorResult result) {
     if (anomalies == null) {
       return;
     }
-    for (final AnomalyDTO anomalyDTO : anomalies) {
+    anomalies.stream().map(this::setAnomalyAuth).forEach(anomalyDTO -> {
```
suvodeep-pyne, line 193: would prefer to have fluent APIs on separate lines
```
anomalies.stream()
    .map(this::setAnomalyAuth)
    .forEach(anomaly -> ...
```

suvodeep-pyne (approved):
LGTM

thirdeye-worker/src/main/java/ai/startree/thirdeye/worker/task/runner/DetectionPipelineTaskRunner.java
```diff
@@ -183,12 +190,12 @@ private void postExecution(final OperatorResult result) {
...
       final Long id = anomalyDao.save(anomalyDTO);
```
cyrilou242, line 194: nitpick: in fluent apis, I try to only use methodReference 
`map(this::doThis)` `map(ClassName::doThisStatic)` or one liners `map(elem -> doThat(elem, 4))`

so in this case it could look like: 
```
anomalies.stream()
  .map(this::setAnomalyAuth)
  .forEach(this::saveAnomaly)
```
the mix of fluent and imperative tends to be less readable imo.
but here it's the limit case, both ways are not that good (the extracted method is shallow)

--> your call.

cyrilou242 (commented):
important: I think we forgot to discuss auth info in child anomalies.
did you have a plan for this?

child anomalies are in anomaly.getChildren()

PR #899, opened 2023-01-24
chore(ui): TE-1216 update luxon version and update usage to reflect major version bump

spham92:
#### Issue(s)

Indirectly related to https://cortexdata.atlassian.net/browse/TE-1216

#### Description

Major version bump luxon from `1.25.0` to `3.2.0`

#### Screenshots

No UI changes


thirdeye-ui/src/app/platform/utils/date-time/date-time.util.ts
```diff
@@ -129,12 +129,19 @@ export const formatDurationV1 = (
 // Returns formatted string representation of date and time
 // For example:
 // MMM DD, YYYY, HH:MM AM/PM
-export const formatDateAndTimeV1 = (date: number): string => {
+export const formatDateAndTimeV1 = (
+    date: number,
+    timezoneOverride?: string
```
alcatraz627, line 134: Just a suggestion, how about making the second parameter an object so it can support other optional params in the future if needed

`{
  timezoneOverride
}?: {
  timezoneOverride?: string
}`

alcatraz627 (approved):

thirdeye-ui/src/app/platform/utils/date-time/date-time.util.ts
spham92: Ahh good suggestion. WIll make the change when we add more options

PR #912, opened 2023-01-27
[postprocessors] remove child generation when merging anomalies

cyrilou242:
- don't generate child anomalies anymore
- all code manipulating child anomalies, and child ids is unchanged for the moment to keep backward compatibility.
- anomaly `avgCurrentValue` is not the average of anomalies when anomalies are merged, it is the value of the first anomaly. This is the current behaviour so this is not changed.


cyrilou242:
Closing - makes rule 3 and 3bis of merge logic for mutability impossible.

PR #913, opened 2023-01-30
feat(ui): TE-1216 enhance timepicker to support displaying and returning selected dates in timezone

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1216

#### Description

Add support to display times in a specific tiemzone if passed. Also support the selected displayed UI buttons return the right timestamp in the respective passed timezone

#### Screenshots

With a timezone passed
image

Without a timezone passed (defaults to local timezone)
image


thirdeye-ui/src/app/utils/time-range/time-range.util.ts
```diff
@@ -199,28 +199,32 @@ export const formatTimeRange = (timeRange: TimeRange): string => {
 
 // Returns formatted string representation of time range duration
 export const formatTimeRangeDuration = (
-    timeRangeDuration: TimeRangeDuration
+    timeRangeDuration: TimeRangeDuration,
+    timezone?: string
```
alcatraz627, line 203: Do you think having this default to `"UTC"` would be better or not?

spham92: I'm leaving the option open to view in the local timezone by having this be undefined

alcatraz627 (approved):

PR #934, opened 2023-02-02
Add AuthorizationConfiguration container for namespace

jacksonargo:
This change adds a container AuthorizationConfiguration to store the namespace and any future fields relating to authorization.

Api resources that support AuthorizationConfiguration:
 - DataSource
 - Dataset
 - AlertTemplate
 - Alert
 - EnumerationItem
 - Anomaly


suvodeep-pyne (approved):
This is great! I just think “auth” is better than “authorization”. The later is just longer IMO without adding much value. other than that LGTM. Note that the name of the field will be hard to change later without migration.

PR #947, opened 2023-02-08
[persistence] Metric to compute system latency of task execution

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1313

#### Description

Introduced metrics to compute system latency of task execution

**Metric** : `detectionTaskLatencyInMillis`
**type** : `CachedGauge`
**Description** : Gives the max latency of a detection task for the requested instant, although the value is cached for 15 seconds. The latency takes into consideration the total time spent by a task in the system when the request is made.

**Metric** : `notificationTaskLatencyInMillis`
**type** : `CachedGauge`
**Description** : Gives the max latency of a notification task for the requested instant, although the value is cached for 15 seconds. The latency takes into consideration the total time spent by a task in the system when the request is made.


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -291,6 +293,41 @@ protected Long loadValue() {
...
+                Predicate.AND(
+                    Predicate.GE("updateTime", taskLatencyWindowStart),
+                    Predicate.LT("updateTime", taskLatencyWindowEnd)
+                )
+            ));
+        final List<TaskDTO> tasks = filter(filter);
+        if(tasks.isEmpty()) {
+          return 0L;
+        }
+        final long maxLatency = tasks.stream().map(task -> {
```
cyrilou242, line 316: My understanding is that this was enough: 
```
select IFNULL(CURRENT_TIMESTAMP() - MIN(create_time), 0) 
from task_index 
where status in ('WAITING', 'RUNNING') and type = 'DETECTION';
```
--> 1 for DETECTION , 1 for NOTIFICATION maybe.

Am I missing cases?

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -67,6 +67,8 @@ public class TaskManagerImpl implements TaskManager {
   private static final String FIND_BY_NAME_ORDER_BY_CREATE_TIME_DESC =
       " WHERE name = :name order by createTime desc limit ";
 
+  private static Timestamp taskLatencyWindowStart = new Timestamp(System.currentTimeMillis());
```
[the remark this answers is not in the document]
shounakmk219: This will be emitted from all pods i.e. coordinator, scheduler, worker and something like max(latency from all pods) will be done to get the value on which alert or any further action is decided.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
shounakmk219: The tasks with fulfilled statuses such as ('COMPLETED', 'FAILED', 'TIMEOUT') , latency of these must also be considered from create_time to last status update_time (usually status change is the only update done on tasks)

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -291,6 +293,41 @@ protected Long loadValue() {
         return count();
       }
     });
+
+    metricRegistry.register("taskLatencyInMillis", new CachedGauge<Long>(15, TimeUnit.SECONDS) {
```
cyrilou242, line 297: I think we should differentiate system latency for notification and detection tasks. 
It'll help when debugging.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -291,6 +293,41 @@ protected Long loadValue() {
...
+      protected Long loadValue() {
+        final List<TaskStatus> pendingStatus = List.of(TaskStatus.WAITING, TaskStatus.RUNNING);
+        final Timestamp taskLatencyWindowEnd = new Timestamp(System.currentTimeMillis());
+        final DaoFilter filter = new DaoFilter()
+            .setPredicate(Predicate.OR(
+                // tasks which are still active in the lifecycle
+                Predicate.IN("status", pendingStatus.toArray()),
+                // tasks which completed(updated with the final status) lifecycle in the window
+                Predicate.AND(
+                    Predicate.GE("updateTime", taskLatencyWindowStart),
```
cyrilou242, line 308: actually still not sure to understand this.
let's discuss on slack

cyrilou242 (changes requested):
let's discuss on slack. 
I think we need to write down formulas before implementing.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
shounakmk219: That's a great point! Will add separate metrics for both type of tasks.

cyrilou242:
LGTM!

cyrilou242 (approved):

PR #948, opened 2023-02-08
Enforce access control for generic entities, dataset onboarding, and anomly feedback

jacksonargo:
Implements access control for:
 - Generic entities api - This would be an access control leak 😓 
 - Dataset onboarding
 - Setting feedback on anomalies


jacksonargo:
Splitting this into 3 different prs

PR #953, opened 2023-02-10
check for read perm when evalutating alert

jacksonargo:
Check for read instead of write when evaluating alerts. This way the UI can show the alert graph for anyone who can read the alert.


suvodeep-pyne (approved):
LGTM

PR #954, opened 2023-02-13
chore(ui): STUI-611 removed hardcoded Slack channel name from release config

joshigaurava:
#### Issue(s)

https://cortexdata.atlassian.net/browse/STUI-611

#### Description

Slack channel name was hardcoded in release configurations. Going forward, it should be picked up in [CI](https://github.com/startreedata/ci-startree-ui/blob/master/.concourse/tasks/thirdeye-ui/publish.task.yaml#L15) (for example).

Reference: https://github.com/juliuscc/semantic-release-slack-bot#environment-variables

If the approach doesn't work, we'll revert to hardcoded value.


opanitch (approved):
looks good 🌮

PR #959, opened 2023-02-15
RCA Investigation gets a copy of the anomaly auth

jacksonargo:
When a new investigation is created, copy authorization config from the parent anomaly.

https://cortexdata.atlassian.net/browse/TE-1328

Tested by creating an investigation from ui w/ access control enabled.

### DIFF
--- a/thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/ApiBeanMapperTest.java
+++ b/thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/ApiBeanMapperTest.java
@@ -17,16 +17,20 @@ import static org.testng.Assert.*;
 
 import ai.startree.thirdeye.spi.api.AlertApi;
 import ai.startree.thirdeye.spi.api.AlertTemplateApi;
+import ai.startree.thirdeye.spi.api.AnomalyApi;
 import ai.startree.thirdeye.spi.api.AuthorizationConfigurationApi;
 import ai.startree.thirdeye.spi.api.DataSourceApi;
 import ai.startree.thirdeye.spi.api.DatasetApi;
 import ai.startree.thirdeye.spi.api.EnumerationItemApi;
+import ai.startree.thirdeye.spi.api.RcaInvestigationApi;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertTemplateDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AnomalyDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.DataSourceDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.DatasetConfigDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.EnumerationItemDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.RcaInvestigationDTO;
 import org.testng.annotations.Test;
 
 // This test validates the mapping between api <-> dto objects.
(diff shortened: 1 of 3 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/resources/RcaInvestigationResource.java
```diff
@@ -48,21 +54,33 @@ public class RcaInvestigationResource extends CrudResource<RcaInvestigationApi,
...
     return rcaInvestigationDTO;
   }
 
   @Override
   protected RcaInvestigationDTO toDto(final RcaInvestigationApi api) {
-    return ApiBeanMapper.toDto(api);
+    final RcaInvestigationDTO dto = ApiBeanMapper.toDto(api);
+
+    // Copy auth from the anomaly.
+    if (dto.getAuth() == null) {
```
suvodeep-pyne, line 76: this is tricky stuff. This is required for all entities. no?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/RcaInvestigationApi.java
```diff
@@ -114,4 +115,13 @@ public RcaInvestigationApi setUpdatedBy(final UserApi updatedBy) {
     this.updatedBy = updatedBy;
     return this;
   }
+
+  public AuthorizationConfigurationApi getAuth() {
```
suvodeep-pyne, line 119: seems like auth is available for all crud entities? Should we place it a level above then?

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/ApiBeanMapperTest.java
```diff
@@ -133,6 +137,46 @@ public void testToEnumerationItemDTO() {
...
+  }
+
+  @Test
+  public void testToAnomalyDTO() {
+    final AnomalyApi api = new AnomalyApi().setAuth(new AuthorizationConfigurationApi().setNamespace(
+        "my-namespace"));
+
+    final AnomalyDTO gotDto = ApiBeanMapper.toDto(api);
+    assertNotNull(gotDto.getAuth());
+    assertEquals(gotDto.getAuth().getNamespace(), "my-namespace");
```
suvodeep-pyne, line 157: would prefer to use assertj APIs in testing. 

it should be available in imports.
```    assertThat(march21Anomalies.stream()
        .anyMatch(a -> a.getEndTime().getTime() == MARCH_23_2020_00H00)).isTrue();
    assertThat(response.getStatus()).isEqualTo(200);

```
doc: https://assertj.github.io/doc/#overview-what-is-assertj

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/RcaInvestigationResource.java
jacksonargo: I'm not sure I understand. What other entities do you have in mind?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/RcaInvestigationApi.java
jacksonargo: The crudApi's implement an interface (unlike the Dto's which extend an abstract class), so I don't think I can move this a level up.

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/RcaInvestigationResource.java
jacksonargo: A safer way to do this might be to have the UI copy the namespace from the anomaly and send it in the POST request, but I'm not sure how quickly that can be done.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/RcaInvestigationApi.java
suvodeep-pyne: We can have getAuth and setAuth just like we have getId and setId in ThirdEyeCrudApi. However, not sure if we are there yet for all the entities. What entities are we not handling today? Else, setting the auth part can be there in CrudResource instead.

suvodeep-pyne: Let's tackle this later though. don't see this as a blocker for this PR

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/RcaInvestigationResource.java
suvodeep-pyne: So, this is used for create/edit investigations right? so, can I set the auth to something and pass in? then that is respected instead of what is set in anomaly?

jacksonargo: Ya. if there's no auth config from the api object, then it receives a copy from the anomaly.

suvodeep-pyne (approved):
LGTM

PR #963, opened 2023-02-16
[persistence] fix anomaly enumerationItem migration logic

cyrilou242:
I think there is a bug in the migration logic that makes enumeration item anomalies messed up when their enumerationItem is migrated.
In the [migrate](https://github.com/startreedata/thirdeye/blob/d5b25e1e14102c53e07139977d849c0a90d510b2/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/EnumerationItemManagerImpl.java#L119) method:
- `from` is an enumerationItem that has no alertId. So it may have been shared between multiple alerts, meaning some anomalies from different alerts can have this enumeration.
- anomalies are fetched with the filter `enumerationItemId=from.Id`
--> **a filter on the alert Id is missing.**
--> anomalies from multiple alerts are migrated to the enumerationItem

--> I think this happened concurrently between multiple alerts. The last migration run by a detection job overrides anomaly enumeration items in others alerts anomalies.
So it's hard to track.


suvodeep-pyne (approved):
LGTM

PR #964, opened 2023-02-16
Introduce root level access + check access for generic entities api

jacksonargo:
This pr adds access control checks to the generic entities api and check for root level perms. 

The generic entities api allows users to view any ThirdEye entity, so access control checks are needed.

The root check requires write permissions within `thirdeye-root` namespace ~and is applied before any other check.~ and is only applied to the entities api.


thirdeye-server/src/main/java/ai/startree/thirdeye/resources/EntityResource.java
```diff
@@ -53,18 +55,23 @@
 public class EntityResource {
 
   private final GenericPojoDao genericPojoDao;
+  private final AuthorizationManager authorizationManager;
 
   @Inject
-  public EntityResource(final GenericPojoDao genericPojoDao) {
+  public EntityResource(final GenericPojoDao genericPojoDao,
```
suvodeep-pyne, line 61: this is another place where I'd prefer `admin` privileges to be able to read internal DB entities.

suvodeep-pyne (dismissed):
LGTM

suvodeep-pyne (approved):

Issue #977, opened 2023-02-18
[kubernetes] Helm chart doesn't support external MySQL instance


dd-willgan:
Currently, the ThirdEye Helm chart also deploys a MySQL instance. While the chart supports backing up the MySQL instance, at our company we would prefer to use an external managed DB service like AWS RDS. Can the chart be updated so that an external URL can be provided?

PR #978, opened 2023-02-18
[kubernetes] Allow MySQL instance to be external

dd-willgan:
#### Issue(s)

- https://github.com/startreedata/thirdeye/issues/977

#### Description

- Use `condition` on the MySQL subchart so that local MySQL doesn't have to be deployed
- Add `.Values.mysql.mysqlUrl` so that external URL can be provided
- Updated configmaps so that `mysqlUrl` is used if provided

#### Screenshots

N/A

#### How to test

From `kubernetes/helm/startree-thirdeye`, I ran the following with different values of `mysql.enabled` and `mysql.mysqlUrl`. Saw working.

```
helm install --dry-run --debug thirdeye . > dry-run.txt
```

### DIFF
--- a/kubernetes/helm/startree-thirdeye/templates/coordinator/configmap.yaml
+++ b/kubernetes/helm/startree-thirdeye/templates/coordinator/configmap.yaml
@@ -40,7 +40,7 @@ data:
           keyStorePassword: {{ .Values.tls.password }}
           keyStoreType: {{ .Values.tls.type }}
         {{- end }}
-  
+
       requestLog:
         appenders:
         - type: console
@@ -59,8 +59,12 @@ data:
 {{ toYaml .Values.auth | indent 6 }}
 
     database:
-      # Assuming a local MySQL server running on the default port 3306
+      # Defaults to k8s cluster local MySQL server running on port 3306, but uses URL if provided
+      {{ if .Values.mysql.mysqlUrl }}
+      url: jdbc:mysql://{{- .Values.mysql.mysqlUrl -}}/{{- .Values.mysql.mysqlDatabase -}}?autoReconnect=true&{{ .Values.config.jdbcParameters }}
+      {{ else }}
       url: jdbc:mysql://{{ include "thirdeye.mysql.fullname" . }}/{{- .Values.mysql.mysqlDatabase -}}?autoReconnect=true&{{ .Values.config.jdbcParameters }}
+      {{ end }}
       user: {{ .Values.mysql.mysqlUser }}
       password: {{ .Values.mysql.mysqlPassword }}
       driver: com.mysql.cj.jdbc.Driver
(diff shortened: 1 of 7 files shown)


mmantri2022 (commented):
Looks good to me. As long as we have ensured no security risk involved.

mmantri2022 (approved):
looks good to me as long as there is no security issues with this and right measure and password protection is taken care.

kubernetes/helm/startree-thirdeye/templates/coordinator/configmap.yaml
```diff
@@ -59,8 +59,12 @@ data:
 {{ toYaml .Values.auth | indent 6 }}
 
     database:
-      # Assuming a local MySQL server running on the default port 3306
+      # Defaults to k8s cluster local MySQL server running on port 3306, but uses URL if provided
+      {{ if .Values.mysql.mysqlUrl }}
```
shounakmk219, line 63: The condition should be on `.Values.mysql.enabled`. That will force user to provide the `mysqlUrl` when internal mysql is disabled. Also avoid the situation of thirdeye connecting to external mysql as well as spinning its own instance if `mysql.mysqlUrl` is also set and `mysql.enabled` is also set to `true`. Let me know your thoughts.

kubernetes/helm/startree-thirdeye/templates/scheduler/scheduler-config.yaml
```diff
@@ -83,8 +83,12 @@ data:
     #      ttl: 60000
 
     database:
-      # Assuming a local MySQL server running on the default port 3306
+      # Defaults to k8s cluster local MySQL server running on port 3306, but uses URL if provided
+      {{ if .Values.mysql.mysqlUrl }}
```
shounakmk219, line 87: Same comment as above

kubernetes/helm/startree-thirdeye/templates/worker/worker-config.yaml
```diff
@@ -83,8 +83,12 @@ data:
     #      ttl: 60000
 
     database:
-      # Assuming a local MySQL server running on the default port 3306
+      # Defaults to k8s cluster local MySQL server running on port 3306, but uses URL if provided
+      {{ if .Values.mysql.mysqlUrl }}
```
shounakmk219, line 87: Same comment as above

shounakmk219 (changes requested):
Nice addition! just few comments. 
You may also want to make changes at the [backup job](https://github.com/startreedata/thirdeye/blob/0d1a8011bbeb580ff5c7fdc86eccfa7344a06b70/kubernetes/helm/startree-thirdeye/templates/mysql/mysql-backup-cronjob.yaml#L51) so that it points to the right mysql instance.

shounakmk219 (approved):

PR #985, opened 2023-02-23
feat(ui): TE-1206 cancel api calls when user leaves some pages

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1206

#### Description

Developers have to explicitly wrap the pages or routers with `CancelAPICallsOnPageUnload` if they wish to cancel API calls on the page or route unload with the react router route definitions.

Some caveats here is that they `key` property is necessary so that the components are unique instance

#### Screenshots

No UI changes

### DIFF
--- a/thirdeye-ui/src/app/components/cancel-api-calls-on-page-unload/cancel-api-calls-on-page-unload.component.tsx
+++ b/thirdeye-ui/src/app/components/cancel-api-calls-on-page-unload/cancel-api-calls-on-page-unload.component.tsx
@@ -18,7 +18,6 @@ import React, { FunctionComponent, useEffect, useState } from "react";
 import { CancelAPICallsOnPageUnloadProps } from "./cancel-api-calls-on-page-unload.interfaces";
 
 /**
- * This has to be wrapped around a page and not a router.
  * The `key` property needs to be set so that this component is unique when
  * used in react router: https://stackoverflow.com/questions/38710250/react-multiple
  * -instances-of-same-component-are-getting-same-state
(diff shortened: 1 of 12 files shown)


thirdeye-ui/src/app/components/cancel-api-calls-on-page-unload/cancel-api-calls-on-page-unload.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * This has to be wrapped around a page and not a router
+ */
+export const CancelAPICallsOnPageUnload: FunctionComponent = ({ children }) => {
+    const [controller] = useState(new AbortController());
+    axios.defaults.signal = controller.signal;
+
+    useEffect(() => {
+        axios.defaults.signal = controller.signal;
```
alcatraz627, line 27: As per the `axios.defaults` documentation, this should be applied globally (instead for just the children components), so would having this wrapper instantiated across multiple routes be any different from just being a wrapper around the root component??

spham92: Yes this is different than having a wrapper at the root component. Having it at the root level will tag any api calls with the same signal across the app.

With this solution, developers have to explicitly wrap the pages they wish to cancel API calls on the page unload with the react router route definitions. This introduces some scoping as some API will be using whatever the abort signal is set to and that's unique and set by this component when its code is run.

alcatraz627 (approved):

PR #1002, opened 2023-03-03
[server] Expose ignore label filter on AnomalyResource

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1362

#### Description

Expose the newly added `ignored` filter as `anomalyLabels.ignore`

**Example request** : `/api/anomalies?anomalyLabels.ignore=false`
This will return all the anomalies which are not ignored


suvodeep-pyne (approved):
LGTM

PR #1008, opened 2023-03-07
simpler template engine

cyrilou242:
- [core] improve json escape in template rendering
- [core] simpler template engine


thirdeye-server/src/test/resources/alertEvaluation.json
```diff
@@ -99,7 +99,5 @@
         }
       ]
     }
-  },
-  "start": "${start}",
-  "end": "${end}"
+  }
```
suvodeep-pyne, line 102: glitch? how does alert evaluation work without start and end?

suvodeep-pyne (approved):
LGTM

This is going deeper into the json serde / jackson libs. there are pros and cons with this. I think this should be ok since the problem at hand is precisely handling substitutions within json which needs to be escaped and is being handled implicitly which is somewhat opaque but nice.

suvodeep-pyne:
Also, the StringTemplateUtils is probably a bad name at this point.

thirdeye-server/src/test/resources/alertEvaluation.json
cyrilou242: The `DetectionPipelineContext` passed to planExecutor$runPipeline contains a detection Interval.
It is consumed by the DataFetcher via the __timeFilterMacro but could be by other ways.

I did this change a long time ago now, I think the main pros at the time were:
- detectionInterval is required by almost all operators, and we must ensure it's the same.
- detectionInterval should be compatible with:
-- timezone
-- granularity bucket rouding
-- completeness delay
-- re-play (recent, for data mutability)

cyrilou242:
> Also, the StringTemplateUtils is probably a bad name at this point.

agree, will rename it in another commit

PR #1018, opened 2023-03-14
[persistence] Metric for notification flow count

shounakmk219:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1346

#### Description

A metric to denote the number of actual subscriptions in the system. A subscription can be described as a unique notification flow which can be denoted by `(alert/enumeration item , notification channel)` tuple. This gives us a high level idea about the volume of agents subscribed to ThirdEye.

Metric : `notificationFlowCountTotal `
Type :  `CachedGauge`


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
```diff
@@ -16,15 +16,36 @@
...
+  public SubscriptionGroupManagerImpl(final GenericPojoDao genericPojoDao,
+      final MetricRegistry metricRegistry) {
     super(SubscriptionGroupDTO.class, genericPojoDao);
+    metricRegistry.register("subscriptionsCountTotal",
+        new CachedGauge<Integer>(15, TimeUnit.MINUTES) {
+      @Override
+      public Integer loadValue() {
+        return findAll().stream()
+            .filter(SubscriptionGroupManagerImpl::isPairPresent)
+            .map(sg -> sg.getAlertAssociations().size() * sg.getSpecs().size())
```
suvodeep-pyne, line 39: This is cool.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
```diff
@@ -16,15 +16,36 @@
...
-  public SubscriptionGroupManagerImpl(GenericPojoDao genericPojoDao) {
+  public SubscriptionGroupManagerImpl(final GenericPojoDao genericPojoDao,
+      final MetricRegistry metricRegistry) {
     super(SubscriptionGroupDTO.class, genericPojoDao);
+    metricRegistry.register("subscriptionsCountTotal",
+        new CachedGauge<Integer>(15, TimeUnit.MINUTES) {
+      @Override
+      public Integer loadValue() {
+        return findAll().stream()
+            .filter(SubscriptionGroupManagerImpl::isPairPresent)
```
suvodeep-pyne, line 38: Instead, just chain the filters. that would be just as readable.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
```diff
@@ -16,15 +16,36 @@
...
+            .map(sg -> sg.getAlertAssociations().size() * sg.getSpecs().size())
+            .reduce(0, Integer::sum);
+      }
+    });
+  }
+  
+  private static boolean isPairPresent(final SubscriptionGroupDTO subscriptionGroup) {
```
suvodeep-pyne, line 45: I understand the function and then the name but the name in itself is super confusing. Like

```
            .filter(SubscriptionGroupManagerImpl::isPairPresent)
```
is not clear to be at all until I read the function code. In this case, I do not see the utility of the method anywhere else and I do not see it as a piece of independent computation.

What would have made sense to me if there was a method to compute the product and return that directly. However, in the current case I think having the logic in the method is way more explicit.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
```diff
@@ -16,15 +16,36 @@
...
   @Inject
-  public SubscriptionGroupManagerImpl(GenericPojoDao genericPojoDao) {
+  public SubscriptionGroupManagerImpl(final GenericPojoDao genericPojoDao,
+      final MetricRegistry metricRegistry) {
     super(SubscriptionGroupDTO.class, genericPojoDao);
+    metricRegistry.register("subscriptionsCountTotal",
+        new CachedGauge<Integer>(15, TimeUnit.MINUTES) {
+      @Override
+      public Integer loadValue() {
+        return findAll().stream()
```
suvodeep-pyne, line 37: if the method definition gets too long after the changes below. pull this out of constructor.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
```diff
@@ -16,15 +16,36 @@
...
 @Singleton
 public class SubscriptionGroupManagerImpl extends
     AbstractManagerImpl<SubscriptionGroupDTO> implements SubscriptionGroupManager {
 
   @Inject
-  public SubscriptionGroupManagerImpl(GenericPojoDao genericPojoDao) {
+  public SubscriptionGroupManagerImpl(final GenericPojoDao genericPojoDao,
+      final MetricRegistry metricRegistry) {
     super(SubscriptionGroupDTO.class, genericPojoDao);
+    metricRegistry.register("subscriptionsCountTotal",
```
suvodeep-pyne, line 33: This name is confusing. If you ask someone from the team what `subscriptionsCountTotal` is, what do you think they would answer?

suvodeep-pyne (changes requested):

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
shounakmk219: Oh I thought the chaining will make it hard to understand why are we having the filters😂 hence I created a separate method so that the filter is more readable. Totally defeated the cause here. Thanks for the feedback, will chain the filters👍

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
shounakmk219: Yeah agree, will dissolve the method in favour of chaining the filters.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
shounakmk219: Noted. I have created separate method named `registerMetrics()` at few places where I felt the constructor is looking too large. If it makes sense we can have this method everywhere by default (wherever metrics are defined in constructor) so that constructor looks clean?

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/SubscriptionGroupManagerImpl.java
shounakmk219: Trust me I had hard time finalizing a name for this metric. Other contender was `notificationFlowCountTotal`. Picked the existing one as we already have a well defined entity `SubscriptionGroup` so atleast no one will confuse the count with `SubscriptionGroup` entity count and at the same time make is obvious that it's related to it somehow.
Any good suggestion is welcomed!

PR #1021, opened 2023-03-22
refactor(ui): TE-1418 alert accuracy fetches in component

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1418

#### Description

When you try to check or even click on an alert while accuracy requests are in flight, you can't since the table keeps re-rendering on every request completion. Refactoring to making the accuracy component make the API request so the table does not re-render.

Also removing the accuracy from enumeration item since that just displays the same data as the header.

#### Screenshots

No visible changes to UI except the removal of the accuracy in enumeration items

image

### DIFF
--- a/thirdeye-ui/src/app/components/alert-accuracy-colored/alert-accuracy-colored.component.tsx
+++ b/thirdeye-ui/src/app/components/alert-accuracy-colored/alert-accuracy-colored.component.tsx
@@ -15,54 +15,138 @@
 
 import { Typography, useTheme } from "@material-ui/core";
 import { capitalize } from "lodash";
-import React, { FunctionComponent } from "react";
+import React, { FunctionComponent, useEffect, useMemo } from "react";
 import { useTranslation } from "react-i18next";
-import { SkeletonV1 } from "../../platform/components";
+import { useInView } from "react-intersection-observer";
+import { SkeletonV1, TooltipV1 } from "../../platform/components";
+import { ActionStatus } from "../../rest/actions.interfaces";
+import { useGetAlertStats } from "../../rest/alerts/alerts.actions";
+import { AnomalyFeedbackType } from "../../rest/dto/anomaly.interfaces";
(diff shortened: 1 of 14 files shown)


thirdeye-ui/src/app/components/alert-accuracy-colored/alert-accuracy-colored.component.tsx
```diff
@@ -15,54 +15,138 @@
...
-        if (!alertStats) {
-            return (
-                renderCustomLoading || (
-                    <SkeletonV1 width={50} {...defaultSkeletonProps} />
-                )
-            );
-        }
+        const { alertStats, getAlertStats, status } = useGetAlertStats();
+
+        const { ref, inView } = useInView({
```
alcatraz627, line 36: This component is used in other places as well (where this loading isn't an issue), so would it be wise to have this data fetching trigger in the component itself?

thirdeye-ui/src/app/components/alert-accuracy-colored/alert-accuracy-colored.component.tsx
```diff
@@ -15,54 +15,138 @@
...
+            triggerOnce: true,
+            delay: 250,
+            threshold: 1,
+        });
+
+        useEffect(() => {
+            if (inView) {
```
alcatraz627, line 43: This component is used in other places as well (where this loading isn't an issue), so would it be wise to have this data fetching trigger in the component itself?

spham92: I don't see an issue having this component fetch the data

alcatraz627: Tested it out. Should be fine.

alcatraz627 (dismissed):

alcatraz627 (approved):

PR #1022, opened 2023-03-22
[Helm] Enable MySQL port to be set when using external

dd-willgan:
#### Issue(s)

- At my company, we only use PostgreSQL instances so a lot of the infrastructure particularly the security group networking rules are tailored for it
- Right now the default 3306 port is not allowed
- It would be easier for us to allow a different port as opposed to trying to modify the networking rules

#### Description

- Allow port to be set if external MySQL is provided, default still 3306

#### How to test

Verified working with different values of

```
helm install --dry-run --debug thirdeye . > dry-run.txt
```


shounakmk219 (approved):
LGTM

PR #1042, opened 2023-04-03
fix(ui): TE-1451 boolean type field take show state of default value if no value set

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1451

#### Description

Have the UI show the state of the default if there is no value set

#### Screenshots

image


thirdeye-ui/src/app/components/alert-wizard-v2/alert-template/alert-template-form-field/form-component-for-template-field/form-component-for-template-field.component.tsx
```diff
@@ -134,7 +134,12 @@ export const FormComponentForTemplateField: FunctionComponent<FormComponentForTe
                     <Grid item>
                         <Switch
                             data-testid={`switch-${propertyKey}`}
-                            defaultChecked={!!value}
+                            defaultChecked={
+                                value === undefined
+                                    ? templateFieldProperty.defaultValue ===
```
joshigaurava, line 139: Not a blocker. Ideally, you would use a platform util to parse a string to boolean. The value could change from "true" to "True" and things will break.

joshigaurava (approved):

PR #1059, opened 2023-04-12
feat(ui): TE-1471 past duration picker

spham92:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1471

#### Description

The component abstracts away building a simple duration string in iso8601 format

#### Screenshots

image

### DIFF
--- a/thirdeye-ui/package.json
+++ b/thirdeye-ui/package.json
@@ -134,6 +134,7 @@
         "html-webpack-plugin": "5.3.2",
         "husky": "7.0.2",
         "identity-obj-proxy": "^3.0.0",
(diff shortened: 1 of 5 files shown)


thirdeye-ui/package.json
```diff
@@ -134,6 +134,7 @@
         "html-webpack-plugin": "5.3.2",
         "husky": "7.0.2",
         "identity-obj-proxy": "^3.0.0",
+        "iso8601-duration": "^2.1.1",
```
joshigaurava, line 137: We should be using absolute versions. Sometimes, breaking changes end up being introduced in minor versions and things break.

PR #1065, opened 2023-04-14
[oauth] Moving oauth logic to plugin code.

suvodeep-pyne:
The current oauth code in thirdeye sits inside the main codebase. moving this out to a plugin to enable pluggability as we add new auth capabilities and decouple auth logic from the server. There is no expected behavior change.

tradeoffs
 - still using the same oauth config class. this should be pluggable in the future
 - removed auth server running metric. this was badly implemented.

#### How to test
Login using oauth should just work as usual.


thirdeye-plugins/thirdeye-oauth/src/main/java/ai/startree/thirdeye/plugins/oauth/ThirdEyePrincipal.java
```diff
@@ -0,0 +1,30 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.plugins.oauth;
+
+import java.security.Principal;
+
+public class ThirdEyePrincipal implements Principal {
```
jacksonargo, line 18: This introduces a second ThirdEyePrincipal class. Is there a way we can merge these classes?

The other class https://github.com/startreedata/thirdeye/pull/1065/files#diff-2b6c0e2ccd2b52eedca9605a31255e929aeeaaf11edfee6e88c450150f92d443L18

thirdeye-plugins/thirdeye-oauth/src/main/java/ai/startree/thirdeye/plugins/oauth/OpenIdConfigurationProviderImpl.java
```diff
@@ -28,7 +29,7 @@
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import javax.annotation.Nullable;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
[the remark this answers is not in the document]
cyrilou242: yes

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/PluginServiceFactory.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package ai.startree.thirdeye.spi;
+
+public interface PluginServiceFactory<PluginServiceT, ContextT> {
```
cyrilou242, line 17: what is the goal of this?

suvodeep-pyne: Goal is to provide structure. If you see, all factories have the same behavior. needs a `name()` and a `build()`. Introducing this now. this is used in both the AuthenticatorFactory and the OpenIdConfigProviderFactory. However, this structure will be usable in all factory classes.

jacksonargo (approved):

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/PluginServiceFactory.java
cyrilou242: ok thanks

PR #1068, opened 2023-04-17
[misc] update pinot version from 0.11.0 to 0.12.1

chouclee:
#### Issue(s)

Pinot version 0.11.0 is not available anymore: https://downloads.apache.org/pinot/

Update the version to 0.12.1 for in quick start scripts.

#### Description

For the newer version of Pinot, pushJobSpec was missing in original configs and the ingestion always fail when uploading the Segment.
Screenshot 2023-04-17 at 2 11 52 PM

Docs on push job spec:
https://docs.pinot.apache.org/v/release-0.12.1/configuration-reference/job-specification#push-job-spec

#### Screenshots

Added this spec and loading data into Pinot works again:
Screenshot 2023-04-17 at 2 11 34 PM

### DIFF
--- a/examples/AdCampaignData/ingestion_job_spec.yaml
+++ b/examples/AdCampaignData/ingestion_job_spec.yaml
@@ -36,3 +36,5 @@ tableSpec:
   tableConfigURI: '${controllerProtocol}://${controllerHost}:${controllerPort}/tables/AdCampaignData'
 pinotClusterSpecs:
   - controllerURI: '${controllerProtocol}://${controllerHost}:${controllerPort}'
+pushJobSpec:
+  pushAttempts: 1
--- a/examples/CleanPageViewsData/ingestion_job_spec.yaml
+++ b/examples/CleanPageViewsData/ingestion_job_spec.yaml
@@ -36,3 +36,5 @@ tableSpec:
   tableConfigURI: '${controllerProtocol}://${controllerHost}:${controllerPort}/tables/CleanPageViewsData'
 pinotClusterSpecs:
   - controllerURI: '${controllerProtocol}://${controllerHost}:${controllerPort}'
+pushJobSpec:
+  pushAttempts: 1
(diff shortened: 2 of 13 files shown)


cyrilou242:
Hey @chouclee, the update to 0.12.1 looks good. 

The `pushJobSpec` issue seems problematic.
Pinot should not introduce backward compatibility issues and in the [doc](https://docs.pinot.apache.org/v/release-0.12.1/configuration-reference/job-specification#push-job-spec) it seems all values inside `pushJobSpec` should have a default value. 
There are many other places with ingestion specs so if we have to change all of them this is an issue.
Let's see with the Pinot team. Can you ask on the #pinot channel?

snleee (approved):
LGTM.

Thank you for calling this issue out. Within StarTree, we don't use ingestion spec and we also are using more recent Pinot version so it's not impacting us. I created OSS issue for calling this out. https://github.com/apache/pinot/issues/10644

chouclee:
@cyrilou242, based on Seunghyun's comment, do you prefer that we add the pushJobSpec now or wait for OSS pinot's fix (then bump the version later on)?

cyrilou242:
@chouclee let's add this fix, I suspect in the future some pinot instance could be in a version with the issue, this could break the [dataloader step](https://github.com/startreedata/thirdeye/blob/master/kubernetes/helm/startree-thirdeye/templates/dataloader/job.yaml). The dataloader steps is a way to load demo data in clients' instances.

What is the reason for overriding the default value? 
```
pushJobSpec:
  pushAttempts: 2
  pushRetryIntervalMillis: 1000
```

What do you think of keeping the current behavior, to keep the change minimal:
```
pushJobSpec:
  pushAttempts: 1
```

For your information once we migrate the local pinot scripts, we also migrate the e2e tests pinot version [here](https://github.com/startreedata/thirdeye/blob/e42aad61f39e9585cb72545d52f7abe50a091d2f/pinot-test-container/src/main/java/org/apache/pinot/testcontainer/PinotContainer.java#L42). This can be done in a separate PR.

chouclee:
@cyrilou242 I just copied over the example config. `pushAttempts` also sgtm to keep the current behavior unchanged. I will make a change.

cyrilou242 (approved):
LGTM thanks! 

Created a ticket to update pinot version in e2e tests
https://cortexdata.atlassian.net/browse/TE-1519

PR #1077, opened 2023-04-18
[server] support new payload fields for AnomalyFeedback

chouclee:
#### Issue(s)

https://cortexdata.atlassian.net/browse/TE-1498

#### Description

Add `created`, `updated`, `updatedBy`, and `reasons` fields for Anomaly Feedback object

#### Screenshots
Screenshot 2023-04-18 at 2 36 39 PM

### DIFF
--- a/thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
+++ b/thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
@@ -16,7 +16,9 @@ package ai.startree.thirdeye.mapper;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import ai.startree.thirdeye.spi.api.AnomalyFeedbackApi;
+import ai.startree.thirdeye.spi.api.UserApi;
 import ai.startree.thirdeye.spi.datalayer.dto.AnomalyFeedbackDTO;
+import ai.startree.thirdeye.spi.detection.AnomalyCause;
 import ai.startree.thirdeye.spi.detection.AnomalyFeedbackType;
 import java.sql.Timestamp;
 import java.time.Instant;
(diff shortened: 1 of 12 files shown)


spham92 (dismissed):
Lgtm

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
```diff
@@ -0,0 +1,88 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import ai.startree.thirdeye.spi.api.AnomalyFeedbackApi;
+import ai.startree.thirdeye.spi.datalayer.dto.AnomalyFeedbackDTO;
+import ai.startree.thirdeye.spi.detection.AnomalyFeedbackType;
+import java.sql.Timestamp;
+import java.time.Instant;
+import org.testng.annotations.Test;
+
+public class AnomalyFeedbackMapperTest {
```
cyrilou242, line 25: I think we should not have tests for mappers. 
These tests catch when a mapping is not done anymore but don't catch when a new field is added but not mapped. 

So they tend to be incomplete and poorly maintained. 
At the API level let's just make sure the API contract is respected. HappyPath tests looks good.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
```diff
@@ -24,6 +27,10 @@ public class AnomalyFeedbackApi {
   private UserApi owner;
   private AnomalyFeedbackType type;
   private String comment;
+  private Date created;
+  private Date updated;
+  private String updatedBy;
```
cyrilou242, line 32: it's not clear in the ticket but let's use the `UserApi owner` instead of introducing a `updatedBy`. 
you can see how it's done in `AlertApi`.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
```diff
@@ -60,4 +67,40 @@ public AnomalyFeedbackApi setComment(final String comment) {
...
+  public String getUpdatedBy() {
+    return updatedBy;
+  }
+
+  public AnomalyFeedbackApi setUpdatedBy(String updatedBy) {
+    this.updatedBy = updatedBy;
+    return this;
+  }
+
+  public List<AnomalyFeedbackReason> getReasons() {
```
cyrilou242, line 98: nitpick: we often use the word _cause_ in the backend code base and in user education, eg root _cause_ analysis. 
Maybe using _causes_ would be more clearer than _reasons_.

What do you think? 
The UI can always change if need be.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedback.java
```diff
@@ -13,6 +13,8 @@
  */
 package ai.startree.thirdeye.spi.detection;
 
+import java.util.List;
+
 public interface AnomalyFeedback {
```
cyrilou242, line 18: this interface is legacy does and not have multiple implems anymore, can you remove it? This can be done as a separate PR. (can be done before or after as you want). Note I think removing it also requires changes in startree-thirdeye.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedbackReason.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.spi.detection;
+
+public enum AnomalyFeedbackReason {
```
cyrilou242, line 16: is this list set in stone? has it been validated? 
Should this really be a plain enum? I fear users will ask for custom values.
Should we plan a free text field for OTHER? 

Eg JET sometimes have DDOS attacks, so I think something like CYBER_ATTACK would be relevant. But maybe it's not.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AnomalyFeedbackDTO.java
```diff
@@ -62,6 +69,15 @@ public AnomalyFeedbackDTO setComment(String comment) {
     return this;
   }
 
+  public List<AnomalyFeedbackReason> getReasons() {
+    return ImmutableList.copyOf(reasons);
```
cyrilou242, line 73: protective copy on get but not on set. This is a bit confusing.
Overall you'll see the DTOs getters/setters are not protective and all fields are non final. This is mostly because of how jackson works.

I'm fine with protective but the setter should be protective then. 
I can't remember all gotchas with jackson, so if you're not sure I think not being protective in DTOs is ok for the moment.

cyrilou242 (changes requested):
The codebase you're working on here is a bit messy! 
Logic looks good, I added a few comments, I think we should take the time to simplify things here.

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
chouclee: `These tests catch when a mapping is not done anymore but don't catch when a new field is added but not mapped.` That's correct, the purpose of these tests are for the first case only. It's impossible for unit test itself to understand whether the missing mapping is by intention or mistake. I used to write unit tests for proto converters (similar to this mapping) but I am ok to remove tests here.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
chouclee: Good suggestion. What about `AnomalyCause`?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AnomalyFeedbackDTO.java
chouclee: Got it. I will remove it.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedback.java
chouclee: Yeah I felt like this interface is unnecessary during the implementation. We can do it in a separate PR.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedbackReason.java
chouclee: User will be provided a multi-checkbox for this field and they are allowed to check multiple reasons (aka causes) when editing the feedback. So it's 1:n relation between feedback and reasons. 

You can check the screenshot in https://cortexdata.atlassian.net/browse/TE-1498

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
chouclee: e.g. `AnomalyFeedbackReason` -> `AnomalyCause` and `getReasons` -> `getCauses`

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
chouclee: ack

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedbackReason.java
cyrilou242: Understood. 
> You can check the screenshot in https://cortexdata.atlassian.net/browse/TE-1498

I'm actually challenging the multi-select. 
Eg in
``` 
HOLIDAY_EFFECT
USER_ERROR
SOFTWARE_FAILURE
ENVIRONMENTAL_FACTOR
FRAUD
PLATFORM_UPGRADE
EXTERNAL_EVENT
OTHER
```

Does it really make sense to have multi-select? 
the possible values look exclusive to me.

Eg if it's a USER_ERROR then it cannot be the other possible values.

I'm fine with multi-select as long as we are aware of at least one case where it's relevant.

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
cyrilou242: Yeah I also used to do tons of proto 2 pojo tests and never found such tests useful.

What do you think @suvodeep-pyne @jacksonargo ?

jacksonargo: I do find the tests useful, especially when debugging the auto-mappers. 

I created a similar unit tests for the dto<->api to verify that the namespace field is properly passed around, because I kept running into cases where the namespace field _wasn't_ mapped for whatever reason.

We can probably use a single test method per Dto/Api instead of method per field, to reduce the boiler-plate code.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyFeedbackApi.java
spham92: Can we expose `createdBy` and `updatedBy`

chouclee: Want to ask another clarification question, if we use `owner` instead, does it mean whoever update the feedback would become the owner of this feedback object? In DTO, we still maintain the `createdBy` and `updatedBy` and `createdBy` would be the user who submitted the original feedback and `updatedBy` would be someone who makes a change.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/detection/AnomalyFeedbackReason.java
chouclee: Had discussion with @spham92 and we decided to support only one cause at this time. So I will make a change on this one.

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
chouclee: Preferred smaller tests because I only covered the cases where mapping annotation is used. It's more readable even though there could more boilerplate code, plus it's generally good for CI tools so all errors can be reported in a single run (otherwise test could stop in the middle if there is any mismatch for tests with multiple assertions). 

But if we also want to test the entire dot <-> api then single dto/api method, I am also ok with a single test covering entire dto -> api, and vice versa. But I extracted out the corner case (ignored fields) as a single test because I want the each test case to cover a single behavior.

chouclee:
Updated feedback entity in anomaly API response:
```
    "feedback": {
        "id": 14618,
        "owner": {
            "principal": "no-auth-user"
        },
        "updatedBy": {
            "principal": "no-auth-user"
        },
        "type": "NOT_ANOMALY",
        "comment": "test updated",
        "created": 1682124869163,
        "updated": 1682125070776,
        "cause": "ENVIRONMENTAL_FACTOR"
    }
```
@spham92 FYI

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/AnomalyFeedbackMapperTest.java
cyrilou242: sounds good

cyrilou242 (approved):
LGTM thanks!

PR #1106, opened 2023-05-23
[core] add timeGroup macros use the exact bucket time column if available

cyrilou242:
introduce `granularityToBucketTimeColumn` - `Map<String, String>` in `DatasetConfigDTO`.
If an exact bucket time column is available for a granularity, TE macros will use it.

Automation of generation and onboarding of such time columns is out of scope of this PR.

Example: 

If a dataset has the following prop: 
```
"timeColumns": [ {
    "name": "timeBucket1hour",
    "granularity": "PT1H",
    "timezone: "UTC"
    },
    {
    "name": "timeBucket5minute",
    "granularity": "PT5M"
    },
]
```
then the alert with granularity `PT1H`/`PT5M` will respectively use `GROUP BY timeBucket1hour`/ `GROUP BY timeBucket5minute`. 
This is faster than using GROUP BY DATETIMECONVERT(...).
For other granularities for which no dedicated time column is available, DATETIMECONVERT is used.

TimeFilter is unchanged to keep using the default time column.

### DIFF
--- a/thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/DatasetConfigDTO.java
+++ b/thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/DatasetConfigDTO.java
@@ -13,6 +13,7 @@
  */
 package ai.startree.thirdeye.spi.datalayer.dto;
 
+import ai.startree.thirdeye.spi.api.TimeColumnApi;
 import ai.startree.thirdeye.spi.datalayer.Templatable;
 import ai.startree.thirdeye.spi.detection.TimeGranularity;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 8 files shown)


thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/DatasetConfigDTO.java
```diff
@@ -41,6 +41,15 @@ public class DatasetConfigDTO extends AbstractDTO {
...
   private List<MetricConfigDTO> metrics;
+  /**
+   * Maintain a map of (granularity, timeBucketColumnName).
+   * The timeBucketColumnName column should contain time buckets in milliseconds that
+   * correspond to the granularity.
+   * This info is used to optimize generated queries.
+   * For instance
+   * {"PT5M": "timeBucket5min", "PT1H": "timeBucket1hour"}
+   */
+  private Map<String, String> granularityToBucketTimeColumn;
```
suvodeep-pyne, line 52: What do you think about this?
```  
private Map<Duration, String> granularityToBucketTimeColumn;
```
I'm trying to see if the schema can enforce the validation.

suvodeep-pyne (approved):
LGTM. This is a great suggestion and implementation!

So the time column still stays the time column and we filter with the standard pinot time column but we skip the DATETIMECONVERT and do a SELECT/GROUP BY with the bucket column. correct?

suvodeep-pyne (dismissed):
LGTM. This is a great suggestion and implementation!

So the time column still stays the time column and we filter with the standard pinot time column but we skip the DATETIMECONVERT and do a SELECT/GROUP BY with the bucket column. correct?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/DatasetConfigDTO.java
suvodeep-pyne: I see your point. yeah. time can be in different granularities and this can expand. I think having a container will be better for the long term. we can add complexity later.

cyrilou242: back to
> I'm trying to see if the schema can enforce the validation.

I saw there is jackson plugin for joda types. I think with this we can introduce schema enforcement with joda types. 
https://github.com/FasterXML/jackson-datatype-joda
I'll try this in an other PR.

PR #1107, opened 2023-05-23
chore(ui): TE-1597 hide alerts after 10 seconds

jacksonargo:
* Alerts automatically go away after 10 seconds.
* First alert is displayed after 0.5s instead of 0.1s. (Gives enough time for an api request to complete before showing 'authentication has been disabled' 😩 )
* ~Only the outermost exception is reported, instead of recursively through the error stack.~

https://github.com/startreedata/thirdeye/assets/7391437/eef1fe18-425b-4615-8bb8-dbd1d0def610


thirdeye-server/src/main/java/ai/startree/thirdeye/core/ExceptionHandler.java
```diff
@@ -92,9 +92,5 @@ private static void populateStatusListApi(final Throwable e, StatusListApi statu
         break;
       }
     }
-
-    if (e.getCause() != null) {
```
suvodeep-pyne, line 96: This actually helps understand the cause. Else, we have to see the logs. We don't have an alternate mechanism

thirdeye-ui/src/app/platform/components/notification-provider-v1/notification-display-v1/notification-display-v1.component.tsx
```diff
@@ -65,6 +65,8 @@ export const NotificationDisplayV1: FunctionComponent<NotificationDisplayV1Props
                     <Snackbar
                         open
                         anchorOrigin={{ vertical: "top", horizontal: "center" }}
+                        autoHideDuration={3000}
```
suvodeep-pyne, line 68: What is the behavior if you have a long list and have them timeout? Also, I think having a 10s timeout would be better. Some of the msg are long and 3s is too short. thoughts?

suvodeep-pyne (changes requested):
Great initiative @jacksonargo in driving this.

thirdeye-ui/src/app/platform/components/notification-provider-v1/notification-display-v1/notification-display-v1.component.tsx
cyrilou242: +1, even 20s is more reasonable to start with I think.

jacksonargo: Ok ya I'll do 10s. The timeout is for each alert serially, so if there are 20 alerts, then first one displays and timeouts out, then next displays and so on.

suvodeep-pyne: thanks for sharing the compounding thing. 10s looks ok to me. We will do a longer term thing when @spham92  is back.

thirdeye-server/src/main/java/ai/startree/thirdeye/core/ExceptionHandler.java
jacksonargo: Ok I'll leave them in. I was going to do one big ui alert that contains all the messages, but the display looks messy so I'll leave it as is.

PR #1141, opened 2023-06-29
[authorization] anomalies and rca inherit auth from parent alert/enum

jacksonargo:
* Stop storing auth info for anomalies ~and rca investigations~
* Anomalies and investigations inherit the auth config from their parent resources.
  * Anomalies -> enum item if exists, else alert
  * ~Investigations -> anomaly's parent~

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
@@ -17,7 +17,6 @@ import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_
 import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_NAME;
 import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_NAMESPACE;
 import static ai.startree.thirdeye.spi.util.SpiUtils.optional;
-import static ai.startree.thirdeye.util.ResourceUtils.ensureExists;
 
 import ai.startree.thirdeye.alert.AlertTemplateRenderer;
 import ai.startree.thirdeye.datalayer.dao.SubEntities;
@@ -32,7 +31,6 @@ import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertTemplateDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AnomalyDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
-import ai.startree.thirdeye.spi.datalayer.dto.RcaInvestigationDTO;
 import com.google.inject.Inject;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -143,45 +141,54 @@ public class AuthorizationManager {
         accessControl.hasAccess(principal.getAuthToken(), ROOT_RESOURCE_ID, AccessType.WRITE);
   }
 
-  public <T extends AbstractDTO> ResourceIdentifier resourceId(final T dto) {
-    final AbstractDTO authParentDto = resolveAuthParentDto(dto);
+  // Returns the resource identifier for a dto.
+  // Null is ok and maps to a default resource id.
+  public ResourceIdentifier resourceId(final AbstractDTO dto) {
+    final var name = optional(dto)
+        .map(AbstractDTO::getId)
+        .map(Objects::toString)
+        .orElse(DEFAULT_NAME);
 
-    return ResourceIdentifier.from(
-        optional(authParentDto.getId()).map(Objects::toString).orElse(DEFAULT_NAME),
-        optional(authParentDto.getAuth())
-            .map(AuthorizationConfigurationDTO::getNamespace).orElse(DEFAULT_NAMESPACE),
-        optional(SubEntities.BEAN_TYPE_MAP.get(authParentDto.getClass()))
-            .map(Objects::toString).orElse(DEFAULT_ENTITY_TYPE));
+    final var namespace = optional(dto)
+        .map(this::resolveAuthParentDto)
+        .map(AbstractDTO::getAuth)
+        .map(AuthorizationConfigurationDTO::getNamespace)
+        .orElse(DEFAULT_NAMESPACE);
+
+    final var entityType = optional(dto)
+        .map(AbstractDTO::getClass)
+        .map(SubEntities.BEAN_TYPE_MAP::get)
+        .map(Objects::toString)
+        .orElse(DEFAULT_ENTITY_TYPE);
+
+    return ResourceIdentifier.from(name, namespace, entityType);
   }
 
   // Resolves the Authorization parent for certain dto classes.
   // AnomalyDTO -> EnumerationItemDTO if exists, else AlertDTO
-  // RcaInvestigationDTO -> AnomalyDTO's parent (recursive call)
   // Everything else -> itself
-  private <T extends AbstractDTO> AbstractDTO resolveAuthParentDto(final T dto) {
+  private AbstractDTO resolveAuthParentDto(final AbstractDTO dto) {
     if (dto instanceof AnomalyDTO) {
       return resolveAuthParentDto((AnomalyDTO) (dto));
-    } else if (dto instanceof RcaInvestigationDTO) {
-      return resolveAuthParentDto((RcaInvestigationDTO) (dto));
     } else {
       return dto;
     }
   }
 
   private AbstractDTO resolveAuthParentDto(final AnomalyDTO dto) {
-    if (dto.getEnumerationItem() != null) {
-      final Long enumItemId = dto.getEnumerationItem().getId();
-      return ensureExists(enumerationItemManager.findById(enumItemId));
+    if (dto == null) {
+      return null;
     }
 
-    final Long alertId = dto.getDetectionConfigId();
-    return ensureExists(alertManager.findById(alertId));
-  }
+    if (dto.getEnumerationItem() != null) {
+      return optional(dto.getEnumerationItem().getId())
+          .map(enumerationItemManager::findById)
+          .orElse(null);
+    }
 
-  private AbstractDTO resolveAuthParentDto(final RcaInvestigationDTO dto) {
-    final Long anomalyId = dto.getAnomaly().getId();
-    final AnomalyDTO anomalyDto = ensureExists(anomalyManager.findById(anomalyId));
-    return resolveAuthParentDto(anomalyDto);
+    return optional(dto.getDetectionConfigId())
+        .map(alertManager::findById)
+        .orElse(null);
   }
 
   private <T extends AbstractDTO> List<ResourceIdentifier> relatedEntities(T entity) {
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
@@ -159,9 +159,9 @@ public class AuthorizationManager {
   // RcaInvestigationDTO -> AnomalyDTO's parent (recursive call)
   // Everything else -> itself
   private <T extends AbstractDTO> AbstractDTO resolveAuthParentDto(final T dto) {
-    if (dto.getClass() == AnomalyDTO.class) {
+    if (dto instanceof AnomalyDTO) {
       return resolveAuthParentDto((AnomalyDTO) (dto));
-    } else if (dto.getClass() == RcaInvestigationDTO.class) {
+    } else if (dto instanceof RcaInvestigationDTO) {
       return resolveAuthParentDto((RcaInvestigationDTO) (dto));
     } else {
       return dto;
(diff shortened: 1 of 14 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -129,15 +144,47 @@ public boolean hasRootAccess(final ThirdEyePrincipal principal) {
...
-        optional(SubEntities.BEAN_TYPE_MAP.get(dto.getClass()))
+        optional(SubEntities.BEAN_TYPE_MAP.get(authParentDto.getClass()))
             .map(Objects::toString).orElse(DEFAULT_ENTITY_TYPE));
   }
 
+  // Resolves the Authorization parent for certain dto classes.
+  // AnomalyDTO -> EnumerationItemDTO if exists, else AlertDTO
+  // RcaInvestigationDTO -> AnomalyDTO's parent (recursive call)
+  // Everything else -> itself
+  private <T extends AbstractDTO> AbstractDTO resolveAuthParentDto(final T dto) {
```
suvodeep-pyne, line 162: why not just use method overloads here? Doing generics first and then if else defeats the purpose.

```
  private  AbstractDTO resolveAuthParentDto(final AnomalyDTO dto);
  private  AbstractDTO resolveAuthParentDto(final RcaInvestigationDTO dto);
...
```

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -129,15 +144,47 @@ public boolean hasRootAccess(final ThirdEyePrincipal principal) {
...
+    if (dto.getClass() == AnomalyDTO.class) {
+      final var anomalyDto = (AnomalyDTO) (dto);
+      if (anomalyDto.getEnumerationItem() != null) {
+        var enumItemId = anomalyDto.getEnumerationItem().getId();
+        return ensureExists(enumerationItemManager.findById(enumItemId));
+      }
+
+      var alertId = anomalyDto.getDetectionConfigId();
+      if (alertId == null) {
```
suvodeep-pyne, line 171: We should fix these tests. Where does anomalies without alert ID exist?

thirdeye-server/src/test/java/ai/startree/thirdeye/resources/AlertResourceTest.java
```diff
@@ -300,7 +306,10 @@ public void testEvaluate_withExistingAlertAndReadAccessToAlertAndPartialAccessTo
         mock(AlertInsightsProvider.class),
         new AuthorizationManager(alertTemplateRenderer,
```
suvodeep-pyne, line 307: seems like we need different instances of `authorizationManager`?

probably better to have `createAuthorizationManager(accessControl)` then? 
Seems like 1 single arg is changing. rest are all the same. 

Also eventually, you may need mocks here.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyApi.java
```diff
@@ -274,12 +273,4 @@ public AnomalyApi setAnomalyLabels(
     return this;
   }
 
-  public AuthorizationConfigurationApi getAuth() {
```
suvodeep-pyne, line 277: Not sure about the API change. How does the UI know that the user can't access this anomaly?
```
example: https://localhost/anomalies/72410/view
```

We may still need to support this as is. And that's probably not a bad idea. thoughts?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/RcaInvestigationApi.java
```diff
@@ -115,13 +114,4 @@ public RcaInvestigationApi setUpdatedBy(final UserApi updatedBy) {
     this.updatedBy = updatedBy;
     return this;
   }
-
-  public AuthorizationConfigurationApi getAuth() {
```
suvodeep-pyne, line 119: Same comment as above. Even if backend doesn't rely on rca entity for auth, we don't need to push the onus on the UI to figure that out.

suvodeep-pyne (changes requested):
Hey @jacksonargo 

Thanks for powering this through. some comments:
- I think Anomaly and RCA Investigation changes should be separate even though similar. 2 different things. 
- I'm not sure about API changes. I think we should retain APIs as is. Also let's double check with @spham92 
- rest are minor comments

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/AnomalyApi.java
jacksonargo: Ya, I think that makes sense. I'm just not sure how to populate the auth stuff from the dto -> api mapper. It seems like I would have to include the authorization manager to resolve the auth config, which feels kinda heavy. 

🤔 It might be worth rethinking the api shape for auth anyways, but I'll leave the auth container in api for now.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
jacksonargo: I added the method overloads. I think I still need the if-else in the generic one, since the caller functions don't have the type info and it needs the type-casting.

jacksonargo:
> What about existing user data. they need to be migrated. no?

Nothing needs to be migrated; any existing auth containers on anomalies will be ignored.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
suvodeep-pyne: Yeah. then you add this. 
```
  private  AbstractDTO resolveAuthParentDto(final AnomalyDTO dto);
  private  AbstractDTO resolveAuthParentDto(final RcaInvestigationDTO dto);
// then you also have this
  private  AbstractDTO resolveAuthParentDto(final AbstractDTO dto);
```

Java should be able to resolve this. or am I missing something here? 🤔

suvodeep-pyne:
Online discussion on items
 - create/edit anomaly resource should not have an auth container
 - toApi operation should populate the auth container
 - integration test for auth, get/set/anomaly, new anomalies generated by the pipeline

jacksonargo:
Added the integration tests in pr https://github.com/startreedata/startree-thirdeye/pull/183

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -129,13 +141,54 @@ public boolean hasRootAccess(final ThirdEyePrincipal principal) {
...
+        .map(Objects::toString)
+        .orElse(DEFAULT_ENTITY_TYPE);
+
+    return ResourceIdentifier.from(name, namespace, entityType);
+  }
+
+  // Resolves the Authorization parent for certain dto classes.
+  // AnomalyDTO -> EnumerationItemDTO if exists, else AlertDTO
+  // Everything else -> itself
+  private AbstractDTO resolveAuthParentDto(final AbstractDTO dto) {
```
suvodeep-pyne, line 170: Yeah this is completely my bad. we need an inheritance hierarchy here. Let's come back to this later. This is ok for now.

thirdeye-server/src/main/java/ai/startree/thirdeye/service/AnomalyService.java
```diff
@@ -78,9 +84,18 @@ protected AnomalyDTO createDto(final ThirdEyePrincipal principal,
...
+    super.validate(api, existing);
+    ensureNull(api.getAuth(), "cannot set auth for anomalies");
+  }
+
   @Override
   protected AnomalyDTO toDto(final AnomalyApi api) {
-    return ApiBeanMapper.toDto(api);
+    final var dto = ApiBeanMapper.toDto(api);
+    final var authId = authorizationManager.resourceId(dto);
+    dto.setAuth(new AuthorizationConfigurationDTO().setNamespace(authId.namespace));
```
suvodeep-pyne, line 97: cosmetic, 

Let's encapsulate to `auth.getNamespace()`

can be done in separate PR.

thirdeye-server/src/test/java/ai/startree/thirdeye/auth/AuthorizationManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    final var authorizationManager = new AuthorizationManager(
+        null, null, null, null, null);
+    final var got = authorizationManager.resourceId(null);
+    assertThat(got.name).isEqualTo(ResourceIdentifier.DEFAULT_NAME);
+    assertThat(got.namespace).isEqualTo(ResourceIdentifier.DEFAULT_NAMESPACE);
+    assertThat(got.entityType).isEqualTo(ResourceIdentifier.DEFAULT_ENTITY_TYPE);
+  }
+
+  @Test
+  public void testResourceId_forAnomalyDto() {
```
suvodeep-pyne, line 40: cosmetic.

'_' underscore is generally frowned upon in Java land. Let's just camelify it.

thirdeye-server/src/test/java/ai/startree/thirdeye/auth/AuthorizationManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AnomalyDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
+import org.testng.annotations.Test;
+
+public class AuthorizationManagerTest {
+
+  @Test
+  public void testResourceId_forNullDto() {
+    final var authorizationManager = new AuthorizationManager(
```
suvodeep-pyne, line 31: shouldn't we just assert? why do we need to support a `null` input?

thirdeye-server/src/test/java/ai/startree/thirdeye/auth/AuthorizationManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    anomalyDto.setId(2L);
+
+    final var authorizationManager = new AuthorizationManager(
+        null, null, alertManager, null, null);
+
+    final var got = authorizationManager.resourceId(anomalyDto);
+    assertThat(got.name).isEqualTo("2");
+    assertThat(got.namespace).isEqualTo("test_namespace");
+    assertThat(got.entityType).isEqualTo("ANOMALY");
+  }
```
suvodeep-pyne, line 58: we should test for enumeration item as well.

thirdeye-server/src/test/java/ai/startree/thirdeye/resources/AnomalyResourceTest.java
```diff
@@ -0,0 +1,103 @@
...
+      final AnomalyManager anomalyManager,
+      final AlertManager alertManager
+      ) {
+    return new AnomalyResource(new AnomalyService(
+        anomalyManager,
+        alertManager,
+        null,
+        new AuthorizationManager(
+            null,
+            AccessControlProvider.alwaysAllow,
```
suvodeep-pyne, line 50: cosmetic. this seems like a static const. convention is to capitalize to `ALWAYS_ALLOW` in that case.

suvodeep-pyne: tip: use the setup method. you might not need different instances of `AnomalyResource`

thirdeye-server/src/test/java/ai/startree/thirdeye/resources/AnomalyResourceTest.java
```diff
@@ -0,0 +1,103 @@
...
+        )
+    ));
+  }
+
+  static ThirdEyePrincipal nobody() {
+    return new ThirdEyePrincipal("nobody", "");
+  }
+
+  @Test
+  void testGet() {
```
suvodeep-pyne, line 63: thinking aloud.

by the look of it, this seems like a test for AnomalyService. It is setup more like an integration test than a unit test and the resource method is really a pass through in this case. thoughts?

suvodeep-pyne (approved):
LGTM overall. Thanks for doing this.

some minor comments. please take a look. if you are pushing stuff out to separate PRs, that's cool too but let's track them in JIRAs.

thirdeye-server/src/test/java/ai/startree/thirdeye/auth/AuthorizationManagerTest.java
jacksonargo: I'm supporting null out of extreme caution in case an alert/enum item doesn't exist for an anomaly. I know these resources _should_ exist, but I don't want to throw a new exception where we otherwise wouldn't have. Ex `GET /api/anomaly/44` can still return complete even if the anomaly has no parent.

thirdeye-server/src/test/java/ai/startree/thirdeye/resources/AnomalyResourceTest.java
jacksonargo: Yes, I agree, this is more of an integration test. I can convert it later.

PR #1145, opened 2023-07-06
[core] fix completenessDelay behaviour

cyrilou242:
only apply completenessDelay when the endTime is bigger than the data watermark

The change is simple but it’s a big design change: /evaluate and /run are not “pure” anymore --> the behaviour depends on the time at which they run.
These parts were already pretty well tested with unit test and e2e scheduling test so I did not add much tests.

### DIFF
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
@@ -96,7 +96,7 @@ public class AlertDetectionIntervalCalculator {
     DateTime correctedEnd = taskEnd;
     // apply delay correction
     final Period delay = getDelay(metadata);
-    final DateTime dataWatermark = new DateTime(chronology).minus(delay);
+    final DateTime dataWatermark = DateTime.now(chronology).minus(delay);
     if (correctedEnd.isAfter(dataWatermark)) {
       correctedEnd = dataWatermark;
       LOG.info(
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
@@ -96,10 +96,13 @@ public class AlertDetectionIntervalCalculator {
     DateTime correctedEnd = taskEnd;
     // apply delay correction
     final Period delay = getDelay(metadata);
-    // if alert has already run: start = lastTimestamp = correctedEnd from last run -> delay was applied
-    // if alert has never run: delay is not important if end-delay > start.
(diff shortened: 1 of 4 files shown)


thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
```diff
@@ -96,10 +96,13 @@ private static Interval getCorrectedInterval(final long alertId, final long task
     DateTime correctedEnd = taskEnd;
     // apply delay correction
     final Period delay = getDelay(metadata);
-    // if alert has already run: start = lastTimestamp = correctedEnd from last run -> delay was applied
-    // if alert has never run: delay is not important if end-delay > start.
-    // only apply delay on start if end-delay > start is not true
-    correctedEnd = correctedEnd.minus(delay);
+    final DateTime dataWatermark = new DateTime(chronology).minus(delay);
```
jacksonargo, line 99: If I understand correctly, this is using the current time when this line is executed.

I think this is fine. We assume that physically, the data within the delay window isn't complete, so there is an inherent dependency on the time of execution and the data that's operated on. Since this interval correction happens outside/before the detection pipeline execution, the pipeline itself is still repeatable `:copium:`.

jacksonargo (dismissed):

thirdeye-core/src/main/java/ai/startree/thirdeye/alert/AlertDetectionIntervalCalculator.java
cyrilou242: yes the pipeline execution is still repeatable

cyrilou242: let me check if there is something clearer like `DateTime.now()`

suvodeep-pyne:
@cyrilou242 LGTM
This seems like a reasonable thing to do. If the end time is beyond the point where data is ready, we are trimming it, else we are not. The same logic should hold in evaluate or the pipeline. no?

cyrilou242:
> @cyrilou242 LGTM This seems like a reasonable thing to do. If the end time is beyond the point where data is ready, we are trimming it else we aren't. It is as if the dataset itself is available till `current time - completenessDelay`.
> 
> I think the same logic should hold in evaluate or the pipeline. no?

@suvodeep-pyne yes, this is applied in evaluate and pipeline because they consume `AlertDetectionIntervalCalculator` intervals. AlertDetectionIntervalCalculator is the single place to compute time roundings/corrections for alert. The detection interval is passed to the evaluate/pipeline as an input.
This way the pipeline does not depend on system time, which is pretty handy for testing/reproducibility/debug.
The fix in AlertInsightsProvider was because of the previously bad design, insights had to be aware of the behaviour of AlertDetectionIntervalCalculator.

PR #1159, opened 2023-07-08
test docker cache

cyrilou242:


cyrilou242:
caching docker images manually is slower than downloading at each run.
main reasons: 
- docker/github interfaces are not good for this, need to perform a manual save/load of image, which is suboptimal
- the save results in a single file. I suspect it is downloaded as such by the cache. Whereas docker download images by layers in parallel.

PR #1164, opened 2023-07-12
[authorization] rca investigations get auth from parent anomaly

jacksonargo:
Akin to https://github.com/startreedata/thirdeye/pull/1141 ,but for investigations.

Investigation auth namespace is copied from the anomaly.

I also removed the anomaly service tests I created in https://github.com/startreedata/thirdeye/pull/1141 and replaced them with integration tests.


suvodeep-pyne (dismissed):
LGTM 👍

PR #1177, opened 2023-07-28
added repository parameter in values.yaml

davizucon:
Fixing sample values.yaml according to chart.

#### Description

according to deployment.yaml that concat : 
  - image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
 and ui:
  - image: "{{ .Values.ui.image.repository }}:{{ .Values.ui.image.tag }}"

### DIFF
--- a/kubernetes/helm/startree-thirdeye/values.yaml
+++ b/kubernetes/helm/startree-thirdeye/values.yaml
@@ -15,12 +15,14 @@
 # Default values for ThirdEye Deployment
 
 image:
+  repository: image
   tag: latest
   pullPolicy: Always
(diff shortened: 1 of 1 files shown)


cyrilou242:
Hey @davizucon thanks! 

You could provide more context about this? 
I need to check again but I think we have a way for helm to throw an error if `repository.image` or `ui.repository.image` is not set. I need to check if we use it in our infra. 

Adding the default value you suggest would break this, and the deployment would still not work because the image value is not valid.

cyrilou242:
no activity - closing

PR #1185, opened 2023-08-11
[server] add textual analysis for related anomalies

cyrilou242:


thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/RelatedAnomaliesAnalysisApi.java
```diff
@@ -0,0 +1,30 @@
+package ai.startree.thirdeye.spi.api;
+
+import java.util.List;
+
+public class RelatedAnomaliesAnalysisApi {
+
+  private List<AnomalyApi> anomalies;
+
+  private TextualAnalysis textualAnalysis;
```
suvodeep-pyne, line 9: I'm thinking we might need this on the anomaly level as well. However, I want to know what you are thinking in terms of the end product.

PR #1199, opened 2023-09-06
cache namespaces

jacksonargo:
https://cortexdata.atlassian.net/browse/TE-1819

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
@@ -15,7 +15,7 @@ package ai.startree.thirdeye.auth;
 
 import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_ENTITY_TYPE;
 import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_NAME;
-import static ai.startree.thirdeye.spi.accessControl.ResourceIdentifier.DEFAULT_NAMESPACE;
+import static ai.startree.thirdeye.spi.util.SpiUtils.optional;
 
 import ai.startree.thirdeye.alert.AlertTemplateRenderer;
 import ai.startree.thirdeye.auth.ThirdEyePrincipal.AuthenticationType;
@@ -29,19 +29,11 @@ import ai.startree.thirdeye.spi.datalayer.bao.EnumerationItemManager;
 import ai.startree.thirdeye.spi.datalayer.dto.AbstractDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertTemplateDTO;
-import ai.startree.thirdeye.spi.datalayer.dto.AnomalyDTO;
-import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
-import ai.startree.thirdeye.spi.datalayer.dto.RcaInvestigationDTO;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import com.google.inject.Inject;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
 import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
@@ -59,15 +51,7 @@ public class AuthorizationManager {
 
   private final AlertTemplateRenderer alertTemplateRenderer;
   private final AccessControl accessControl;
-  private final AlertManager alertManager;
-  private final EnumerationItemManager enumerationItemManager;
-  private final AnomalyManager anomalyManager;
-
-  // cache resource id's to their respective auth namespaces
-  private final Cache<Long, String> namespaceCache = CacheBuilder.newBuilder()
-      .maximumSize(2048)
-      .expireAfterAccess(5, TimeUnit.SECONDS)
-      .build();
+  private final NamespaceResolver namespaceResolver;
 
   @Inject
   public AuthorizationManager(
@@ -78,9 +62,7 @@ public class AuthorizationManager {
       final AnomalyManager anomalyManager) {
     this.alertTemplateRenderer = alertTemplateRenderer;
     this.accessControl = accessControl;
-    this.alertManager = alertManager;
-    this.enumerationItemManager = enumerationItemManager;
-    this.anomalyManager = anomalyManager;
+    this.namespaceResolver = new NamespaceResolver(alertManager, enumerationItemManager, anomalyManager);
   }
 
   public <T extends AbstractDTO> void ensureCanCreate(final ThirdEyePrincipal principal,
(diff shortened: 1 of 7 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -145,18 +155,16 @@ public boolean hasRootAccess(final ThirdEyePrincipal principal) {
   // Returns the resource identifier for a dto.
   // Null is ok and maps to a default resource id.
   public ResourceIdentifier resourceId(final AbstractDTO dto) {
-    final var name = optional(dto)
+    final var name = Optional.ofNullable(dto)
```
suvodeep-pyne, line 158: This is the same thing. didn't understand this

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -58,6 +62,12 @@ public class AuthorizationManager {
   private final EnumerationItemManager enumerationItemManager;
   private final AnomalyManager anomalyManager;
 
+  // cache resource id's to their respective auth namespaces
+  private final Cache<Long, String> namespaceCache = CacheBuilder.newBuilder()
```
suvodeep-pyne, line 66: Did you check out `LoadingCache`? It's probably a subclass of this.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -58,6 +62,12 @@ public class AuthorizationManager {
...
+      .maximumSize(2048)
+      .expireAfterAccess(5, TimeUnit.SECONDS)
```
suvodeep-pyne, line 68: How often are people changing policies?

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -166,41 +174,60 @@ public ResourceIdentifier resourceId(final AbstractDTO dto) {
...
         .map(alertManager::findById)
-        .orElse(null);
+        .map(this::resolveAuthNamespaceFromCache)
+        .orElse(DEFAULT_NAMESPACE);
   }
 
-  private AbstractDTO resolveAuthParentDto(final RcaInvestigationDTO dto) {
-    return optional(dto)
+  private String resolveRcaNamespace(final RcaInvestigationDTO dto) {
+    return Optional.of(dto)
```
suvodeep-pyne, line 225: same comment for `optional()` as above.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -58,6 +62,12 @@ public class AuthorizationManager {
   private final EnumerationItemManager enumerationItemManager;
   private final AnomalyManager anomalyManager;
 
+  // cache resource id's to their respective auth namespaces
+  private final Cache<Long, String> namespaceCache = CacheBuilder.newBuilder()
+      .maximumSize(2048)
```
suvodeep-pyne, line 67: Given its long vs string. this should be a small map. How did we decide on size 2k?

suvodeep-pyne (commented):
I think we should move namespace resolution to its dedicated class now.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
jacksonargo: 🤔 I'm not sure. I think I'd rather have the update/create methods add to the cache instead of expiring like this.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
jacksonargo: Oh ya it is, just feels a little unnecessary to use the wrapper. I can change it back.

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
jacksonargo: I just picked a number. Haven't settled on the _right_ number yet. Since the map should be small, would it make sense to cache namespaces for all entities up front?

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
jacksonargo: Ah ya I can use that. The difference is in the `get` method: Cache.get needs the loading function, whereas LoadingCache.get uses loading function from builder.

jacksonargo: Ended up using regular cahce, since the loader part depends on the entity type (alert loaded by alert manager, anomaly loaded by anomaly manager)

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
```diff
@@ -67,9 +61,7 @@ public AuthorizationManager(
       final AnomalyManager anomalyManager) {
     this.alertTemplateRenderer = alertTemplateRenderer;
     this.accessControl = accessControl;
-    this.alertManager = alertManager;
-    this.enumerationItemManager = enumerationItemManager;
-    this.anomalyManager = anomalyManager;
+    this.namespaceResolver = new NamespaceResolver(alertManager, enumerationItemManager, anomalyManager);
```
suvodeep-pyne, line 64: we should inject this dependency

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/NamespaceResolver.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+public class NamespaceResolver {
+
+  private final AlertManager alertManager;
+  private final EnumerationItemManager enumerationItemManager;
+  private final AnomalyManager anomalyManager;
+
+  private static final Cache<Long, String> namespaceCache = CacheBuilder.newBuilder()
```
suvodeep-pyne, line 37: Let's not do static?

thirdeye-server/src/main/java/ai/startree/thirdeye/service/CrudService.java
```diff
@@ -122,6 +122,7 @@ public List<ApiT> createMultiple(final ThirdEyePrincipal principal,
         .map(api -> createDto(principal, api))
         .map(dto -> createGateKeeper(principal, dto))
         .peek(dtoManager::save)
+        .peek(dto -> authorizationManager.invalidateCache())
```
suvodeep-pyne, line 125: This needs to be done once.

thirdeye-server/src/main/java/ai/startree/thirdeye/service/CrudService.java
```diff
@@ -135,6 +136,7 @@ public List<ApiT> editMultiple(final ThirdEyePrincipal principal,
         .map(o -> updateDto(principal, o))
         .peek(dtoManager::update)
         .peek(this::postUpdate)
+        .peek(dto -> authorizationManager.invalidateCache())
```
suvodeep-pyne, line 139: same as above

thirdeye-server/src/main/java/ai/startree/thirdeye/auth/NamespaceResolver.java
```diff
@@ -0,0 +1,129 @@
...
+        .orElse(DEFAULT_NAMESPACE);
+  }
+
+  private String getEnumerationItemNamespaceById(final long id) {
+    try {
+      return namespaceCache.get(id, () ->
+          optional(enumerationItemManager.findById(id))
+              .map(this::getNamespaceFromAuth)
+              .orElse(DEFAULT_NAMESPACE));
+    } catch (ExecutionException e) {
```
suvodeep-pyne, line 96: remove catch and escalate instead.

suvodeep-pyne (approved):
LGTM. please address comments and merge

PR #1247, opened 2023-11-29
[mvn] remove helix dependency

cyrilou242:


cyrilou242:
ignoring the snyk failure. 
this PR is part of upgrading dependencies for better security and is making progress, even if there was a potential regression on the zookeeper version

PR #1295, opened 2024-01-17
[server] fix enumeration item namespace inheritance

cyrilou242:
- add failing test
- [server] fix enumerationItem namespace inheritance


suvodeep-pyne (commented):
LGTM. We can iterate on the resolution flow at a later time.

PR #1344, opened 2024-03-01
[worker] don't save results of detection job if alert was edited

cyrilou242:


cyrilou242:
not merging. We will rework this potential concurrency in a different way.

PR #1347, opened 2024-03-04
te 2165 add get active namespace

cyrilou242:
for the moment toy implementation to check the limits of the design


cyrilou242:
no need for this toy implem anymore

PR #1389, opened 2024-04-18
fix(ui): TE-2114 ask user if feedback should re-tune alert

alcatraz627:
#### Issue(s)

[TE-2114](https://startree.atlassian.net/browse/TE-2114)

#### Description

- Add a checkbox to the anomaly feedback modal asking the user if they want this feedback to tune the alert.

#### Screenshots

- Submit Feedback
Screenshot 2024-04-18 at 23 13 32

- Edit Feedback
Screenshot 2024-04-18 at 23 13 45


chouclee (approved):

thirdeye-ui/src/app/rest/dto/anomaly.interfaces.ts
```diff
@@ -61,6 +61,7 @@ export interface AnomalyFeedback {
         principal: string;
     };
     updated?: number;
+    updateAlertConfiguration?: "ASYNC";
```
chouclee, line 64: nit. you can use enum if you think this will have different modes. If no, you can simply use boolean?

alcatraz627: As of now, this option can just take one value, in the future it would support multiple modes. Having just one mode right now, I added that as a the only specific type value it can take.

PR #1391, opened 2024-04-19
fix(ui): TE-2134 correct sensitivity range in filters page

harshilvelotio:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2134

#### Description

- Fixed the sensitivity slider values to respect the backend restrictions of `Sensitivity must be in [-26.2, 14.9]`

#### Screenshots

image

#### How to test
- Adjust the sensitivity slider on the Filters page to "High" and check if 14.9 is sent as the sensitivity value when reloading the chart

### DIFF
--- a/thirdeye-ui/src/app/components/alert-wizard-v3/anomalies-filter-panel/anomalies-filter-panel.component.tsx
+++ b/thirdeye-ui/src/app/components/alert-wizard-v3/anomalies-filter-panel/anomalies-filter-panel.component.tsx
@@ -107,22 +107,22 @@ export const AnomaliesFilterPanel: FunctionComponent<AnomaliesFilterPanelProps> 
                                                 }
                                                 marks={[
                                                     {
-                                                        value: -14,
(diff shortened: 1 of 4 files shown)


cyrilou242:
Hey @harshilvelotio the bounds must be defined/used somewhere else to, because they are set correctly when doing: 
- create alert
- simpler alert - create
- cleanPageViewsDataset, views, NEXT
- Select StarTree-ETS
see secreenshot:
Screenshot 2024-04-19 at 10 23 46


So I'd prefer if the same values were used from a shared constant. 
From what I could see the values are -26, -6, 14 --> you should be able to find them in the codebase.

cyrilou242 (approved):
LGTM for the min/max values thanks @harshilvelotio!  
I'll let thirdeye-ui team approve for the code.

alcatraz627 (approved):

alcatraz627:
LGTM!

PR #1396, opened 2024-04-23
fix(ui): TE-2212 search by subscription group for anomalies bug fix

harshilvelotio:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2212
#### Description
- For subscription groups with dx alerts that have no `enumerationItem` attached to them, the filtering algorithm on the UI had to be changed
- The algorithm now supports filtering across all types of alerts

#### Screenshots
image

#### How to test
- Try filtering through various subscription groups on anomalies list page with subscription to simple and dx alerts with and without `enumerationItem` attached to them

### DIFF
--- a/thirdeye-ui/src/app/pages/anomalies-all-page/anomalies-all-page.component.tsx
+++ b/thirdeye-ui/src/app/pages/anomalies-all-page/anomalies-all-page.component.tsx
@@ -145,7 +145,11 @@ export const AnomaliesAllPage: FunctionComponent = () => {
                             subscriptionGroup.alertAssociations.forEach(
                                 (association) => {
                                     associatedAlertsEnumerationKey.push(
-                                        `${association.alert.id}-${association?.enumerationItem?.id}`
+                                        `${association.alert.id}${
+                                            association?.enumerationItem?.id
+                                                ? `-${association?.enumerationItem?.id}`
+                                                : ""
+                                        }`
                                     );
                                 }
                             );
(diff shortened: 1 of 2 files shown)


thirdeye-ui/src/app/pages/anomalies-all-page/anomalies-all-page.component.tsx
```diff
@@ -145,7 +145,11 @@ export const AnomaliesAllPage: FunctionComponent = () => {
                             subscriptionGroup.alertAssociations.forEach(
                                 (association) => {
                                     associatedAlertsEnumerationKey.push(
-                                        `${association.alert.id}-${association?.enumerationItem?.id}`
```
cyrilou242, line 148: I'm still not clear on the issue, 

why 
```
`${association.alert.id}-${association?.enumerationItem?.id}`
```

And 
```
`${anomaly.alert.id}-${anomaly?.enumerationItem?.id}`
```

do not return the same result when `enumerationItem` or `enumerationItem.id` is not set? 

my understanding is it should in both cases return 
```
3-undefined
```
(assuming alert.id = 3)

is this caused by some typing in typescript?

harshilvelotio: What's happening is that let's say if `${association.alert.id}-${association?.enumerationItem?.id}` is `3-undefined` but the anomaly has an `anomaly?.enumerationItem?.id` then that turns out to be `3-5` and hence the equality check fails and hence filtering of anomalies fails

So now I've altered the check this way

1. If `association.enumerationItem.id` doesn't exist (is undefined) - then filter anomalies by
```
`association.alert.id` === `anomaly.alert.id`
```

2. If `association.enumerationItem.id` does exist - then filter anomalies by 
```
${association.alert.id}-${association?.enumerationItem?.id}` === `${anomaly.alert.id}-${anomaly?.enumerationItem?.id}`
```

cyrilou242: oh ok then why keep the string key to then split it later https://github.com/startreedata/thirdeye/pull/1396/files#diff-3d39297a650808712607bb657c6c6a118f391337639166fedc492c76ed509959R164 ? 

Could you just use an object instead? 

```
associatedAlertsEnumerationKey
  .push({"alertId": association.alert.id, 
         "enumId": association?.enumerationItem?.id}`
```

Then 
```
associatedAlertsEnumerationKey
  .some(
       (k: <the_object>) => {
           k.alertId === anomaly.alert.id && (k.enumId is undefined || k.enumId === anomaly.enumerationItem?.id)
```


In java we would go even further by introducing a Matcher, which is the key object + the function match(), 
This way you only have to 
```
associatedAlertsEnumerationKey.some((k: AssociationMatcher) => {k.match(anomaly))
```
no opinion in javascript.

harshilvelotio: Good option let me refactor

cyrilou242 (approved):
LGTM thanks!

PR #1400, opened 2024-04-25
fix(ui): TE-2056 better error message for pinot broker exception

harshilvelotio:
#### Issue(s)
https://startree.atlassian.net/browse/TE-2056

#### Description
- If there's a Pinot broker exception when running queries while investigating an anomaly, the error messaging needed to be improved

#### Screenshots
image

#### How to test
- Try to simulate an error where there's a Pinot Client Exception raised when running a query


thirdeye-ui/src/app/components/rca/top-contributors-section/top-contributors-section.component.tsx
```diff
@@ -83,7 +84,15 @@ export const TopContributorsSection: FunctionComponent<TopContributorsSectionPro
         useEffect(() => {
             notifyIfErrors(
                 anomalyDimensionAnalysisReqStatus,
-                errorMessages,
+                generateCustomErrorMessage(errorMessages, [
+                    {
+                        criteriaFunction: (error) =>
+                            error
+                                .toLowerCase()
+                                .includes("pinotclientexception"),
```
cyrilou242, line 92: this will get broken when we change error management on the backend side. 
the frontend should only match on 
the `code` field which is part of the API.

Eg: 
```
code: "ERR_ALERT_PIPELINE_EXECUTION"
```

Let me check if it's easy to introduce a new code error for Pinot exceptions.

cyrilou242 (changes requested):

thirdeye-ui/src/app/locale/languages/en-us.json
```diff
@@ -714,6 +715,7 @@
         "property-value-placeholder": "Type and press enter to add values",
         "provide-an-optional-description-for-this-alert": "Provide an optional description for this alert",
         "query-example": "e.g. country='IN' AND device = ‘iPhone’",
+        "query-timeout-error": "Raise a ticket with support team, to fix query performance issues",
```
cyrilou242, line 718: I think it should be prefixed with `pinot-`. Also it'll make it easier to add more PINOT error code and pinot error messages in the same place.  
We also have MySQL queries.

thirdeye-ui/src/app/locale/languages/en-us.json
```diff
@@ -705,6 +705,7 @@
         "our-new-feature-sets-up-your-alert-with-the-parameters": "Our new feature sets up your alert with the parameters that best fit your data. (currently in beta)",
         "page-not-found": "Requested page could not be found",
         "percentage-rule-algorithm-description": "Compare current timeseries to a baseline. If the percentage change is above a certain threshold, detects an anomaly.",
+        "pinot-broker-error": "Raise a ticket with support team to fix Pinot broker listening issues.",
```
cyrilou242, line 708: Add something like: 
"Pinot query failed"? 
Also we should still be able to see the query, it really helps when debugging. 
I think the simplest way would be to have a `<details>` that make it possible to show the `message`  fields returned by the backend.

thirdeye-ui/src/app/components/rca/top-contributors-section/top-contributors-section.component.tsx
harshilvelotio: Makes sense! Can you send the exact code to match 2 kinds of errors
- Error with Pinot broker
- Timeout error (assuming this exception is thrown by the backend and not rely on browser timeout)
?

cyrilou242: There are no such code yet. You can mark the ticket as blocked and mention me. I won't be able to work on this today.

harshilvelotio:
Closing this since a fresh PR will be created accommodating overall error messaging feedback

PR #1415, opened 2024-05-06
fix(ui): TE-2256 fix the datasets and datasource api signatures

harshilvelotio:
#### Issue(s)
https://startree.atlassian.net/browse/TE-2256

#### Description

- Fixed the API call signatures to use datasets ID than name

#### How to test


cyrilou242:
LGTM thanks. 

Will merge to deploy quickly

PR #1421, opened 2024-05-08
fix(ui): TE-2278 add a column and sort by percent change in the top contributors table

harshilvelotio:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2278

#### Description

- Added a column for "Change %" to the "Top Contributors" table when investigating an anomaly

#### Screenshots
image

#### How to test
- Try toggling the sort for the "Change %" column of the "Top Contributors" section when investigating an anomaly

### DIFF
--- a/thirdeye-ui/src/app/components/rca/top-contributors-table/top-contributors-table.component.tsx
+++ b/thirdeye-ui/src/app/components/rca/top-contributors-table/top-contributors-table.component.tsx
@@ -18,9 +18,12 @@ import {
     TableCell,
     TableHead,
     TableRow,
+    withStyles,
 } from "@material-ui/core";
+import { KeyboardArrowDown, KeyboardArrowUp } from "@material-ui/icons";
 import React, { FunctionComponent, useMemo } from "react";
 import { useTranslation } from "react-i18next";
+import { AnomalyDimensionAnalysisMetricRow } from "../../../rest/dto/rca.interfaces";
 import {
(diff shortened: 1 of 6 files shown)


thirdeye-ui/src/app/components/rca/top-contributors-table/top-contributors-table.component.tsx
```diff
@@ -76,6 +133,17 @@ export const TopContributorsTable: FunctionComponent<TopContributorsTableProps>
                         <TableCell>
                             <strong>{t("label.impact-percentage")}</strong>
                         </TableCell>
+                        <SortableTableHeader onClick={toggleSort}>
+                            <strong>{t("label.change-percentage")}</strong>
```
alcatraz627, line 137: While I understand the use case here, it is better to avoid nesting ternary operators like this. 

- Can use a value returned by `useMemo` or something similar with a list of branching if-return statements, so it is easier to understand what is happening
- Alternatively, can have multiple rendering clauses, like
```jsx
{changePercentSort === null && (
  <></> //Is this empty fragment needed for proper rendering?
)}
{changePercentSort === TopContributorsTableChangePercentSort.ASC && (
  <KeyboardArrowUp />
)}
{changePercentSort === TopContributorsTableChangePercentSort.DESC && (
  <KeyboardArrowDown />
)}
```

Not a blocker comment, can fix this later if the PR is urgent.

alcatraz627 (approved):

PR #1423, opened 2024-05-09
fix(ui): TE-2236 fix condition for showing dimension for a dx alert

harshilvelotio:
#### Issue(s)
https://startree.atlassian.net/browse/TE-2236

#### Description

- Fixed the condition that shows the dimension charts for dx alerts when viewing an alert

#### Screenshots
image

#### How to test
- Try viewing multiple DX alerts and check if the dimension charts are visible


cyrilou242:
LGTM thanks

PR #1442, opened 2024-05-24
Missing .Values.ui.resources

dd-willgan:
#### Description

This is defined in values.yaml but not being used in the deployment spec


suvodeep-pyne (approved):
LGTM

PR #1443, opened 2024-05-27
fix(ui): TE-2319 fix insights endpoint on view anomaly new UI

harshilvelotio:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2319

#### Description

- Added a null safety check to make sure the alert id of an anomaly is sent correctly to the /insights endpoint, customer reported that he's been seeing some 400 errors (malformed request) which means that in some cases the alert id is not attached correctly

#### Screenshots

image

#### How to test

- Keep switching between old and new UI for the view anomaly page


cyrilou242:
Hey thanks for this. 
So what happens if the alert.id is indeed null?

cyrilou242:
Overall looking at https://startree.atlassian.net/browse/TE-2319 I'd like to push back until we have the exact root cause and we can reproduce.
I added some additional logs in the backend before my holidays. The customer will install it this week. We should be able to get more info about the root cause.

alcatraz627 (approved):

PR #1518, opened 2024-08-06
[mvn] Upgrade jackson to latest version

anshul98ks123:
#### Issue(s)
[SECURITY - upgrade jackson to latest version](https://startree.atlassian.net/browse/TE-2383)

#### Description
Upgrading Jackson to the latest version to fix [DoS vulnerability](https://github.com/startreedata/thirdeye/pull/1504)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
     <guava.version>32.1.3-jre</guava.version>
     <guice.version>5.1.0</guice.version>
     <hsqldb.version>2.7.1</hsqldb.version>
-    <jackson.version>2.13.5</jackson.version>
+    <jackson.version>2.17.2</jackson.version>
     <jetty.version>9.4.48.v20220622</jetty.version>
     <jodatime.version>2.12.5</jodatime.version>
     <maven-assembly-plugin.version>3.3.0</maven-assembly-plugin.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -934,6 +935,12 @@
         <artifactId>hsqldb</artifactId>
         <version>${hsqldb.version}</version>
       </dependency>
+      <dependency>
```
cyrilou242, line 938: any reason for introducing this snakeyaml dependency?

cyrilou242 (changes requested):

pom.xml
cyrilou242: what is a bit confusing is that snakeyaml should be managed by jackson as a transitive dependency.
We should not declare it. 
It was indeed upgraded to 2.2 in the version 2.16: 
https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.16
so we should not have to do it manually. 

I suspect there is a conflicting version of this dependency somewhere else. It'd be nice to understand where it comes from.
Need to make sure upgrading to 2.2 does not break something else.

cyrilou242: you may want to use 

```
./mvnw dependency:tree
```

to understand what happens

anshul98ks123: Even the ThirdEyeDebugServer fails locally if i bump up the Jackson version and not add snakeyaml as a dependency. I removed the test scope and then the server runs as expected.

anshul98ks123:
Based on offline discussion, Force upgrading snakeyaml version since other way would require upgrading multiple other dependencies

cyrilou242 (approved):
LGTM thanks!

PR #1520, opened 2024-08-09
[server] Fix rca heatmap aggregation fn logic

anshul98ks123:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2382

#### Description

Aggregation function for RCA heatmap is resolved by following logic currently:
1. If custom RCA aggregation fn is provided, use it
2. if metric does not exists, default to COUNT 
3. else, use default aggregation fn of the metric

In cases 2 and 3, alert aggregation function should be used for RCA heatmap if custom rca aggregation fn is not present in template properties

This PR introduces changes to include a new metadata property in alert template - detectionAggregationFunction, which will be set to the alert aggregation function. It is being used to decide aggregation fn for rca heatmap if custom rca aggregation fn is not provided

#### Test
Scenario: An alert which is based on MAX aggregation function, but the metric's default aggregation fn is SUM

Before:
In absence of custom rca aggregation fn, we are using metric's default aggregation (SUM) for rca heatmap
screencapture-localhost-7004-root-cause-analysis-v2-anomaly-56389-investigate-what-where-page-2024-08-09-13_21_22

After:
In absence of custom rca aggregation fn, we will now use alert aggregation fn (MAX) for rca heatmap
screencapture-localhost-7004-root-cause-analysis-v2-anomaly-56376-investigate-what-where-page-2024-08-09-13_04_10

Note: 
this is a template breaking change for RCA.
All templates need to be updated. (see change in commit).

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
@@ -34,6 +34,7 @@ import ai.startree.thirdeye.spi.datalayer.dto.AlertDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertMetadataDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AlertTemplateDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.AnomalyDTO;
+import ai.startree.thirdeye.spi.datalayer.dto.AuthorizationConfigurationDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.DataSourceDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.DatasetConfigDTO;
 import ai.startree.thirdeye.spi.datalayer.dto.EnumerationItemDTO;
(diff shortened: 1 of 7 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
```diff
@@ -145,6 +145,9 @@ public RcaInfo getRcaInfo(final AnomalyDTO anomalyDTO)
           .setName(metricName)
```
cyrilou242, line 145: I think we can skip fetching from the persistence layer altogether.
The long term plan is to use this metric abstraction less and less and eventually remove it. (it has other issues, like ThirdEye needs to maintain the metrics in sync with the table schema).

So I think we can just construct the metricConfigDTO manually likes it's done line 144. 
```
metricConfigDTO = new MetricConfigDTO()
           .setDataset(datasetName)
          .setName(metricName)
          .setDefaultAggFunction(
         optional(metadataMetricDTO.getDefaultAggregationFunction).orElse(
         optional(alertMetadataDto.getDetectionAggregationFunction()).orElse(COUNT));
```

cyrilou242 (changes requested):

thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
cyrilou242: note this will change the logic to fetch the dataset below, we will need to fetch by name, namespace in all cases (this is fine)

anshul98ks123:
Makes sense. Addressed the comments

thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
```diff
@@ -127,37 +127,22 @@ public RcaInfo getRcaInfo(final AnomalyDTO anomalyDTO)
...
-      datasetConfigDto = datasetDAO.findById(metricConfigDTO.getDatasetConfig().getId());
-    } else {
-      LOG.warn(
-          "Running on a legacy metrics that does not contain the id of its parent dataset. Finding dataset by name and namespace.");
-      datasetConfigDto = datasetDAO.findByNameAndNamespaceOrUnsetNamespace(
-          metricConfigDTO.getDataset(),
-          alertNamespace);
-    }
+    final String alertNamespace = alertDTO.namespace();
+    final DatasetConfigDTO datasetConfigDto = datasetDAO.findByNameAndNamespaceOrUnsetNamespace(
```
cyrilou242, line 140: please use findUniqueByNameAndNamespace 
(I think)

cyrilou242 (changes requested):

anshul98ks123:
Addressed review comments

cyrilou242 (approved):
LGTM thanks!

PR #1522, opened 2024-08-09
[server] Fix rca heatmap aggregation fn logic

anshul98ks123:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2382

#### Description

Aggregation function for RCA heatmap is resolved by following logic currently:
1. If custom RCA aggregation fn is provided, use it
2. if metric does not exists, default to COUNT 
3. else, use default aggregation fn of the metric

In cases 2 and 3, alert aggregation function should be used for RCA heatmap if custom rca aggregation fn is not present in template properties

This PR introduces changes to include a new metadata property in alert template - detectionAggregationFunction, which will be set to the alert aggregation function. It is being used to decide aggregation fn for rca heatmap if custom rca aggregation fn is not provided

#### Test
Scenario: An alert which is based on MAX aggregation function, but the metric's default aggregation fn is SUM

Before:
In absence of custom rca aggregation fn, we are using metric's default aggregation (SUM) for rca heatmap
screencapture-localhost-7004-root-cause-analysis-v2-anomaly-56389-investigate-what-where-page-2024-08-09-13_21_22

After:
In absence of custom rca aggregation fn, we will now use alert aggregation fn (MAX) for rca heatmap
screencapture-localhost-7004-root-cause-analysis-v2-anomaly-56376-investigate-what-where-page-2024-08-09-13_04_10

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
@@ -131,20 +131,10 @@ public class RcaInfoFetcher {
     // we get the namespace from the alert and inject it in all persistence access - authz is not performed - the consumer of this method should ensure authz on the anomaly
     final String alertNamespace = alertDTO.namespace();
     MetricConfigDTO metricConfigDTO = metricDAO.findBy(metricName, datasetName, alertNamespace);
-    if (metricConfigDTO == null) {
-      LOG.warn("Could not find metric {} for dataset {}. Building a custom metric for RCA.",
-          metricName, datasetName);
-      String metricAggFunction = metadataMetricDTO.getDefaultAggFunction();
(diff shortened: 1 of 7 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/rca/RcaInfoFetcher.java
```diff
@@ -131,20 +131,10 @@ public RcaInfo getRcaInfo(final AnomalyDTO anomalyDTO)
     // we get the namespace from the alert and inject it in all persistence access - authz is not performed - the consumer of this method should ensure authz on the anomaly
     final String alertNamespace = alertDTO.namespace();
     MetricConfigDTO metricConfigDTO = metricDAO.findBy(metricName, datasetName, alertNamespace);
```
cyrilou242, line 133: I think we should not fetch the metric config from the persistence layer anymore and construct it manually.

Will put more details in the other PR

PR #1531, opened 2024-08-16
fix(ui): TE-2411 avoid passing alertId in evaluate call when updating alert

nalin-patidar:
… alert

#### Issue(s)

https://startree.atlassian.net/browse/TE-2411

#### Description

When updating an alert which has enumeration items, we do not have to pass the alertId to the evaluate call 

#### Screenshots

https://github.com/user-attachments/assets/25c002b5-6eff-422c-89da-ddb0d98559b7

#### How to test

1. Go to any alert which has enumeration items.
2. Open the edit alert page.
3. Update any alert property.
4. Click on Load chart preview.
5. See in the evaluate api call in network tab that alertId is not being passed in the payload


thirdeye-ui/src/app/components/alert-wizard-v3/preview-chart/enumeration-items-table/enumeration-item-row/enumeration-item-row.component.tsx
```diff
@@ -67,19 +68,22 @@ export const EnumerationItemRow: FunctionComponent<EnumerationItemRowProps> = ({
     const [timeSeriesData, setTimeSeriesData] = useState<any>();
     const [timeSeriesExpandedData, setTimeSeriesExpandedData] = useState<any>();
 
+    const alertParams = cloneDeep(alert);
+    delete (alertParams as any).id;
```
ghoshnirmalya, line 72: It would be better if we don't use `any`.

ghoshnirmalya (approved):

PR #1535, opened 2024-08-29
chore(ui): TE-2388 added build check and test run

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2388

#### Description

Added build check and test run for pushed to PRs other then master


.github/workflows/fe-build-and-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+  build-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Pull Repository
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            thirdeye-ui
+            pom.xml
```
cyrilou242, line 21: curious on why you need pom.xml and .mvn ? 
I remember some time ago we made the java build tools and the frontend tools totally separated.

.github/workflows/fe-build-and-test.yaml
```diff
@@ -0,0 +1,36 @@
+name: FE Test and Build
+
+on:
+  push:
+    branches:
+      - '!master'
```
cyrilou242, line 6: do we have tests running in another CI in master as of today (can't see it here: https://github.com/startreedata/thirdeye/blob/e35fbd77a91af0655751a2370ceb4b822252c292/.github/workflows/publish-frontend.yaml#L47)? 
if not I think we should also run in master - we will have tests for the edge cases of: 
- user forcing a PR to be merged before PR tests pass
- user pushing to master directly
the CI should be cheap so no need to over optimize here, I'd rather be on the safe side.

.github/workflows/fe-build-and-test.yaml
nalin-patidar: I was referencing the publish action and in that its used for distribution step. And yes we don't need it in this action. Will remove it.

.github/workflows/fe-build-and-test.yaml
```diff
@@ -0,0 +1,34 @@
...
+            thirdeye-ui
+
+      - name: Install node and npm
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+          cache: "npm"
+          cache-dependency-path: './thirdeye-ui/package-lock.json'
+
+      - name: Install dependencies, run test, and build
```
ghoshnirmalya, line 29: Wouldn't it be better to do this in separate steps? Can you please check if, when you create separate steps, the installed `node_modules` gets cached in the GitHub action's artifacts?

ghoshnirmalya (approved):

PR #1546, opened 2024-09-04
[mvn] bump dropwizard to version 3.0.8

cyrilou242:
we can ignore the license check from Snyk - I added the license mention. 
I think I need to do another operation in Snyk.

TODO: 
- [ ] check ThirdEyeServer manually
- [ ] check ThirdEyeDebugServer manually
- [ ] check swagger manually
- [ ] check integration with FrontEnd manually
- [ ] Follow up in enterprise TE


thirdeye-server/pom.xml
```diff
@@ -104,6 +100,12 @@
       <artifactId>converter-jackson</artifactId>
     </dependency>
 
+    <!-- support for joda types in server configuration -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-joda</artifactId>
+    </dependency>
```
anshul98ks123, line 107: was support for joda types removed in version upgrade, that we are adding it explicitly? 
or is it something new we are introducing for possible future need?

cyrilou242: yes.
in dw 2.1.x: 
dw-core --> depends on [dw-jackson](https://mvnrepository.com/artifact/io.dropwizard/dropwizard-jackson/2.1.12) --> depends on jackson-datatype-joda

but in dw-jackson 3.x.y, joda is not a dependency anymore. 
https://mvnrepository.com/artifact/io.dropwizard/dropwizard-jackson/3.0.8

this is most likely because joda is getting deprecated in favor of `java.time` jsr310 
https://github.com/JodaOrg/joda-time
previously java did not have a good time library, but now it has. 
dw-jackson 3.x.y only imports [jackson-datatype-jsr310](https://mvnrepository.com/artifact/com.fasterxml.jackson.datatype/jackson-datatype-jsr310) now

we will eventually have to migrate thirdeye to `java.time` and remove joda but it's a significant effort

cyrilou242: so I have to importe the joda jackson module manually here: 
https://github.com/startreedata/thirdeye/pull/1546/files#diff-90441890a03fc38b378d1624561b9374963d6ddd48e04ef073b9ff264a8398c4R112

anshul98ks123 (approved):

PR #1549, opened 2024-09-10
chore(ui): TE-2414 upgrade axios version

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2414

#### Description

Upgrade axios version to fix snyk vulnerability

### DIFF
--- a/thirdeye-ui/package.json
+++ b/thirdeye-ui/package.json
@@ -64,7 +64,7 @@
         "@tanstack/react-query": "^4.29.19",
         "@tippyjs/react": "^4.2.6",
         "@visx/visx": "3.1.2",
-        "axios": "1.3.4",
+        "axios": "^1.7.4",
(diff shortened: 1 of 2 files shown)


thirdeye-ui/package.json
```diff
@@ -64,7 +64,7 @@
         "@tanstack/react-query": "^4.29.19",
         "@tippyjs/react": "^4.2.6",
         "@visx/visx": "3.1.2",
-        "axios": "1.3.4",
+        "axios": "^1.7.4",
```
jayeshchoudhary, line 67: use exact version without ^

nalin-patidar: We are using the packacge.lock file and checking it out in source control too. Also we are using npm CI everywhere.
So we should be good with carat I think.

esakkiraj (approved):

PR #1577, opened 2024-10-01
[server] Support time configuration on namespace level

anshul98ks123:
#### Issue(s)

[Time configuration - introduce namespace configuration entity](https://startree.atlassian.net/browse/TE-2447)

#### Description
Currently, the time configuration is set statically in the server configuration file
As we move to interstellar setup, this needs to be handled at a per namespace level

This PR Introduces a new configuration entity per namespace - _Namespace Configuration_ 
The namespace configuration entity will store all namespace level customizations (time configuration, etc)  

When a user/internal client/flow (Alert creation, notification builder, etc) fetches the configuration entity for a namespace, if the configuration entity does not exist, ThirdEye creates one automatically, else it returns the existing configuration for that namespace

#### Testing
Tested locally by running the Server and UI
1. Onboarded datasets and created alerts across multiple namespaces
2. Verified that new table was created and new entries were populated, one for each namespace (including null namespace)
3. Verified swagger APIs are working as expected

| API | Expectation |
| --- | --- |
|GET` /api/namespace-configuration/`|should return namespace configuration as per namespace header value. If there is no namespace configuration for provided namespace, then it will create new one and return.|
|PUT `/api/namespace-configuration/`|should allow updating only the time configuration (timezone/datetimepattern/minimumOnboardingStartTime) of namespace configuration entity. Will be used by thirdeye-ui in future to allow users to update namespace configuration.|
|POST `/api/namespace-configuration/reset/`|should reset existing namespace configuration to default values if it exists, else create new one from default values.|


thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    timezone                       varchar(100) NOT NULL,
+    date_time_pattern              varchar(100) NOT NULL,
+    minimum_onboarding_start_time  bigint(20)   NOT NULL,
+    create_time                    timestamp,
+    version                        int(10),
+    update_time                    timestamp DEFAULT CURRENT_TIMESTAMP,
+    json_val                       text
+    ) ENGINE = InnoDB;
+CREATE INDEX namespace_configuration_id_idx ON namespace_configuration_entity (id);
+CREATE INDEX namespace_configuration_namespace_idx ON namespace_configuration_entity (namespace);
```
cyrilou242, line 14: let's set a uniqueness constraint instead
```
ALTER TABLE namespace_configuration_entity ADD UNIQUE unique_namespace (namespace);
```
a uniqueness constraint is also an index under the hood in MySQL

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE INDEX namespace_configuration_create_time_idx ON namespace_configuration_entity (create_time);
```
cyrilou242, line 15: what's the intent of these timestamp indexes ? 
we will only fetch by `namespace` in most cases right? or maybe a full scan for some statistics?

cyrilou242 (changes requested):

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/NamespaceConfigurationApi.java
```diff
@@ -0,0 +1,50 @@
...
+ * the License.
+ */
+package ai.startree.thirdeye.spi.api;
+
+public class NamespaceConfigurationApi implements ThirdEyeCrudApi<NamespaceConfigurationApi> {
+
+  private Long id;
+  private AuthorizationConfigurationApi auth;
+
+  TimeConfigurationApi timeConfiguration;
```
cyrilou242, line 21: private ?

thirdeye-spi/pom.xml
```diff
@@ -98,6 +98,14 @@
       <artifactId>checker-qual</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
cyrilou242, line 101: following edits to this PR guice can be removed right?

thirdeye-notification/src/main/java/ai/startree/thirdeye/notification/NotificationEventsBuilder.java
```diff
@@ -64,7 +66,13 @@ public NotificationEventsBuilder(final EventManager eventDao,
   /**
    * TODO cyril - logic is duplicated (and not iso) with what RcaRelatedService#getRelatedEvents provides
    */
-  public List<EventApi> getRelatedEvents(final Collection<AnomalyDTO> anomalies) {
+  public List<EventApi> getRelatedEvents(final Collection<AnomalyDTO> anomalies,
+      SubscriptionGroupDTO subscriptionGroup) {
+    final String currentNamespace = subscriptionGroup.namespace();
```
cyrilou242, line 71: I think the namespace can be obtained from an anomaly

if (anomalies.isEmpty()) {
  return Collections.emptyList()
}
final String namespace = anomalies.get(0).namespace;

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
```diff
@@ -0,0 +1,195 @@
...
+  public long count() {
+    return dao.count();
+  }
+
+  @Override
+  public long count(final Predicate predicate) {
+    return dao.count(predicate);
+  }
+
+  private TimeConfigurationDTO getDefaultTimeConfiguration() {
```
cyrilou242, line 189: this will break existing environments that have a custom time configuration.
to avoid breaking changes I think the default should be obtained from the Time configuration. 

once we are sure every users has migrated we remove the time configuration and hardcode it here

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/config/NamespaceServerConfigurationManager.java
```diff
@@ -0,0 +1,38 @@
...
+package ai.startree.thirdeye.spi.config;
+
+import ai.startree.thirdeye.spi.datalayer.bao.NamespaceConfigurationManager;
+import ai.startree.thirdeye.spi.datalayer.dto.NamespaceConfigurationDTO;
+import com.google.inject.Inject;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NamespaceServerConfigurationManager {
```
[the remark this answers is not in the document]
anshul98ks123: Cleaned up. No longer needed

cyrilou242 (changes requested):

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
anshul98ks123: Makes sense. Adding

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
anshul98ks123: not needed for now ig. removing

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
anshul98ks123: Ah, right
this is a critical consideration. Let me fix this

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/NamespaceConfigurationApi.java
anshul98ks123: Done

thirdeye-spi/pom.xml
anshul98ks123: Yes, thanks for catching this
done

thirdeye-notification/src/main/java/ai/startree/thirdeye/notification/NotificationEventsBuilder.java
anshul98ks123: makes sense, we don't need to pass subscription group just for this

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
anshul98ks123: Done

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/NamespaceConfigurationResource.java
```diff
@@ -0,0 +1,145 @@
...
+@OpenAPIDefinition(security = {
+    @SecurityRequirement(name = "oauth"),
+    @SecurityRequirement(name = Constants.NAMESPACE_SECURITY)
+})
+@SecurityScheme(name = "oauth", type = SecuritySchemeType.APIKEY, in = SecuritySchemeIn.HEADER, paramName = HttpHeaders.AUTHORIZATION)
+@SecurityScheme(name = Constants.NAMESPACE_SECURITY, type = SecuritySchemeType.APIKEY, in = SecuritySchemeIn.HEADER, paramName = Constants.NAMESPACE_HTTP_HEADER)
+@Singleton
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class NamespaceConfigurationResource extends CrudResource<NamespaceConfigurationApi,
```
cyrilou242, line 62: Maybe it's ok to not extend CrudResource in this case.

1. We only need a GET and a PUT.
GET should always return a single element so enforcing it at the API level will force to limit potential implementation errors (eg namespace filter not working as expected)
For internal use we may want to perform stats on configuration across multiple namespaces but it's a very limited use case (not sure we will ever do this)

2. the problem when we override the CRUD ressources endpoints is the endpoints are still visible in swagger API. This makes the API harder to understand.

3. this will make this class way shorter

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/NamespaceConfigurationResource.java
```diff
@@ -0,0 +1,145 @@
...
+  // Overridden to disable endpoint
+  @Override
+  @PUT
+  @Operation(summary = "", hidden = true)
+  @Timed(percentiles = {0.5, 0.75, 0.90, 0.95, 0.98, 0.99, 0.999})
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response editMultiple(
+      @Parameter(hidden = true) @Auth final ThirdEyeServerPrincipal principal,
+      final List<NamespaceConfigurationApi> list) {
+    throw new UnsupportedOperationException();
```
cyrilou242, line 119: we do need a PUT to edit right? see comment above

cyrilou242 (changes requested):
comments on the API level

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS namespace_configuration_entity
+(
+    id                             bigint(20) PRIMARY KEY AUTO_INCREMENT,
+    namespace                      varchar(200),
+    timezone                       varchar(100) NOT NULL,
```
cyrilou242, line 5: if we don't use JOOQ, is there a use case for flattening timezone, date_time_pattern, minimum_onboarding_start_time? 
we usually flatten to add indexes (btw it's not really required anymore in MySQL 8, we could index json fields directly I think)

```
id
namespace
create_time
update_time
version
```

my understanding is the only query patterns will be: 
```
select where namespace= 
update where namespace = 
select where id=
update where id=
```

cyrilou242 (changes requested):

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
anshul98ks123: sorry, i don't understand this
are you suggesting to have an inner table called TimeConfiguration and add that as a column instead of individual fields directly? if yes, do we have any such example that i can refer?

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/NamespaceConfigurationResource.java
anshul98ks123: Yes, i was planning to add it later. but it's a small change. Adding

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/NamespaceConfigurationResource.java
anshul98ks123: Makes sense. simplifying both files

anshul98ks123: Done

thirdeye-persistence/src/main/resources/db/migration/mysql/V1_307_0__add_namespace_configuration_entity_table.sql
anshul98ks123: nvm, understood 
Removed flatten out fields since they're not needed for now

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    return createNewNamespaceConfiguration(namespace);
+  }
+
+  private NamespaceConfigurationDTO fetchExistingNamespaceConfiguration(String namespace) {
+    final DaoFilter daoFilter = new DaoFilter().setPredicate(Predicate.EQ(
+        "namespace", namespace));
+    final List<NamespaceConfigurationDTO> results = filter(daoFilter);
+    if (results != null && !results.isEmpty()) {
```
cyrilou242, line 110: let's ensure the list only has a single element before .get(0).
if it has multiple elements, the db is in an inconsistent state (or somehow the query above is not valid anymore) and we should not return any configuration

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/ApiResource.java
```diff
@@ -158,4 +161,9 @@ public TaskResource getTaskResource() {
   public UiResource getUIResource() {
     return uiResource;
   }
+
+  @Path("namespace-configuration")
```
cyrilou242, line 165: NamespaceResource is using the name `workspaces` so let's also use `workspace` instead of `namespace` here
https://github.com/startreedata/thirdeye/blob/61b2be6ded9ae74b95c3cfc3f63dc6e7b1e9c619/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/ApiResource.java#L122

thirdeye-server/src/main/java/ai/startree/thirdeye/service/AlertService.java
```diff
@@ -127,6 +124,11 @@ protected void deleteDto(final AlertDTO dto) {
 
   @Override
   protected void prepareCreatedDto(final ThirdEyePrincipal principal, final AlertDTO dto) {
+    final String currentNamespace = authorizationManager.currentNamespace(principal);
```
cyrilou242, line 127: in prepareCreatedDto, the alert dto must have its namespace already set correctly. 
so you can use 
```
dto.namespace()
```
directly

thirdeye-server/src/main/java/ai/startree/thirdeye/service/AlertService.java
```diff
@@ -388,6 +390,11 @@ private void deleteAssociatedRcaInvestigations(final Long alertId) {
   }
 
   private long minimumLastTimestamp(final ThirdEyePrincipal principal, final AlertDTO dto) {
+    final String currentNamespace = authorizationManager.currentNamespace(principal);
```
cyrilou242, line 393: dto must have its namespace set when this method is called, so let's use 
```
dto.namespace()
```

cyrilou242 (changes requested):

thirdeye-core/src/test/java/ai/startree/thirdeye/mapper/ApiBeanMapperTest.java
```diff
@@ -150,4 +156,46 @@ public void testToAuthorizationConfigurationDTO() {
     final AuthorizationConfigurationDTO gotDto = ApiBeanMapper.toAuthorizationConfigurationDTO(api);
     assertThat(gotDto.getNamespace()).isEqualTo("my-namespace");
   }
+
```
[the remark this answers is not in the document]
anshul98ks123: done

thirdeye-notification/src/main/java/ai/startree/thirdeye/notification/NotificationEventsBuilder.java
```diff
@@ -125,4 +132,11 @@ private List<EventDTO> getHolidayEvents(final DateTime start, final DateTime end
         EventType.HOLIDAY.name(), startTimeWithOffsetMillis, endTimeWithOffsetMillis);
     return allEventsBetweenTimeRange;
   }
+
+  private String deduceNamespace(Collection<AnomalyDTO> anomalies) {
+    if (anomalies.isEmpty()) {
+      return null;
```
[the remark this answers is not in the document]
anshul98ks123: makes sense. simplified

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
```diff
@@ -0,0 +1,246 @@
...
+    return dao.update(entities);
+  }
+
+  @Override
+  public NamespaceConfigurationDTO findById(final Long id) {
+    return dao.get(id);
+  }
+
+  @Override
+  public @Nullable NamespaceConfigurationDTO findUniqueByNameAndNamespace(
```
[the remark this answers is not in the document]
anshul98ks123: done. removed unnecessary methods

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
```diff
@@ -0,0 +1,246 @@
...
+    return dao.delete(ids);
+  }
+
+  @Override
+  public int deleteByPredicate(final Predicate predicate) {
+    return dao.deleteByPredicate(predicate);
+  }
+
+  @Override
+  public int deleteRecordsOlderThanDays(final int days) {
```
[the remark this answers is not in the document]
anshul98ks123: done

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
anshul98ks123: done. it should throw exception in an invalid state

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/ApiResource.java
anshul98ks123: done

thirdeye-server/src/main/java/ai/startree/thirdeye/service/AlertService.java
anshul98ks123: done

thirdeye-server/src/main/java/ai/startree/thirdeye/service/AlertService.java
anshul98ks123: done

anshul98ks123:
> LGTM! :shipit: Great work! I have a few little changes in mind related to the DAO/manager apis, I'll create another PR and let you review.

can we merge this PR? or you'll adding new commit in this same PR?

cyrilou242:
> > LGTM! :shipit: Great work! I have a few little changes in mind related to the DAO/manager apis, I'll create another PR and let you review.
> 
> can we merge this PR? or you'll adding new commit in this same PR?

let's merge I'll suggest changes in another PR

PR #1584, opened 2024-10-08
[spi] Add generateDatasourceConnection to ThirdEyeAuthorizer interface

anshul98ks123:
#### Issue(s)

#### Description

Adding generateDatasourceConnection method to ThirdEyeAuthorizer interface
The purpose of this method is to provide placeholder to implement one's own custom datasource connection

One can override the default implementation by using/implementing ThirdEyeAuthorizerPlugin

This method is not being used anywhere currently
It will be integrated into Authorization Manager in upcoming PR to be exposed as a new API

#### Testing
Existing Tests and CI

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ThirdEyeAuthorizerProvider.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ThirdEyeAuthorizerProvider.java
@@ -15,6 +15,7 @@ package ai.startree.thirdeye.auth;
 
 import static com.google.common.base.Preconditions.checkState;
 
+import ai.startree.thirdeye.spi.api.DataSourceApi;
 import ai.startree.thirdeye.spi.auth.AccessType;
 import ai.startree.thirdeye.spi.auth.ResourceIdentifier;
 import ai.startree.thirdeye.spi.auth.ThirdEyeAuthorizer;
@@ -89,6 +90,11 @@ public class ThirdEyeAuthorizerProvider implements ThirdEyeAuthorizer {
     return getAccessControl().listNamespaces(principal);
   }
 
+  @Override
+  public DataSourceApi generateDatasourceConnection(final ThirdEyePrincipal principal) {
+    return getAccessControl().generateDatasourceConnection(principal);
+  }
+
   @VisibleForTesting
   public static class AlwaysAllowAuthorizer implements ThirdEyeAuthorizer {
 
@@ -120,6 +126,11 @@ public class ThirdEyeAuthorizerProvider implements ThirdEyeAuthorizer {
       }
       return LIST_OF_NULL;
     }
+
+    @Override
+    public DataSourceApi generateDatasourceConnection(final ThirdEyePrincipal principal) {
+      return null;
+    }
   }
 
   private static class AlwaysDenyAuthorizer implements ThirdEyeAuthorizer {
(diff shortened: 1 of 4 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/auth/ThirdEyeAuthorizerProvider.java
```diff
@@ -89,6 +90,11 @@ public boolean authorize(final ThirdEyePrincipal principal, final ResourceIdenti
     return getAccessControl().listNamespaces(principal);
   }
 
+  @Override
+  public DataSourceApi generateDatasourceConnection(final ThirdEyePrincipal principal) {
```
cyrilou242, line 94: let's mark the return value as `@Nullable`

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/auth/ThirdEyeAuthorizer.java
```diff
@@ -37,6 +38,14 @@ boolean authorize(
   // if the implementation does not support multiple namespace, please return a list with the value null Collections.singletonList(null)
   @NonNull List<String> listNamespaces(final ThirdEyePrincipal principal);
 
+  /**
+   * Generates the datasource connection configuration given principal
+   *
+   * @param principal the principal to generate datasource for
+   * @return DataSourceApi
+   */
+  DataSourceApi generateDatasourceConnection(final ThirdEyePrincipal principal);
```
cyrilou242, line 47: this interface is implemented by our customers in custom plugins we don't control.
so we need a default behavior to avoid a breaking change.
I think `return null` is fine.

cyrilou242, line 47: let's add `@Nullable`to the return value and explain the meaning of null in the doc. 

smthing like
```
Returns null if there is no recommended datasource configuration. 
```

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/auth/ThirdEyeAuthorizer.java
```diff
@@ -37,6 +38,14 @@ boolean authorize(
   // if the implementation does not support multiple namespace, please return a list with the value null Collections.singletonList(null)
   @NonNull List<String> listNamespaces(final ThirdEyePrincipal principal);
 
+  /**
+   * Generates the datasource connection configuration given principal
```
cyrilou242, line 42: maybe let's clarify the intent for our future selves. Especially given it's not the best design as discussed this morning.

```
Generates a recommended datasource configuration to use in the ThirdEye instance.
Does not create the datasource.

Example: an authorizer plugin could:
1. set up a service account to connect to Pinot 
2. return a valid datasource configuration that uses the service account just created.
```

cyrilou242 (changes requested):

cyrilou242 (approved):
LGTM!

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/auth/ThirdEyeAuthorizer.java
```diff
@@ -37,6 +39,19 @@ boolean authorize(
...
+   * Does not create the datasource.
+   *
+   * Example: an authorizer plugin could:
+   * 1. set up a service account to connect to Pinot
+   * 2. return a valid datasource configuration that uses the service account just created.
+   *
+   * @param principal the principal to generate datasource for
+   * @return DataSourceApi -> Returns null if there is no recommended datasource configuration
+   */
+  @Nullable DataSourceApi generateDatasourceConnection(final ThirdEyePrincipal principal);
```
rohit1576, line 53: generateDatasourceConnection <> generateDataSourceConnection

PR #1585, opened 2024-10-08
[server] Add recommend API to datasource resource

anshul98ks123:
#### Issue(s)

#### Description

Adding new API `api/data-sources/recommend`
It will be used in datasource onboarding UI flow to generate recommended datasource connection config 
(importantly for Shared SaaS setup)

Depends on https://github.com/startreedata/thirdeye/pull/1584

#### Testing
UTs and
Sanity swagger test (should return 200 response with null body due to default implementation)

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java
@@ -133,6 +133,16 @@ public class DataSourceResource extends CrudResource<DataSourceApi, DataSourceDT
     return respondOk(onboarded);
   }
 
+  @POST
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+  @Path("recommend")
+  @Timed(percentiles = {0.5, 0.75, 0.90, 0.95, 0.98, 0.99, 0.999})
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response recommendConfiguration(
+      @Parameter(hidden = true) @Auth ThirdEyeServerPrincipal principal) {
+    return respondOk(dataSourceService.recommend(principal));
+  }
+
   @DELETE
   @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
   @Path("offboard-all")
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java
@@ -134,7 +134,6 @@ public class DataSourceResource extends CrudResource<DataSourceApi, DataSourceDT
   }
 
   @POST
-  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
   @Path("recommend")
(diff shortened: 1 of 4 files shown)


thirdeye-server/src/main/java/ai/startree/thirdeye/service/DataSourceService.java
```diff
@@ -175,4 +177,8 @@ public boolean validate(final ThirdEyePrincipal principal, final long id) {
     authorizationManager.ensureCanRead(principal, dataSourceDto);
     return dataSourceCache.getDataSource(dataSourceDto).validate();
   }
+
+  public DataSourceApi recommend(final ThirdEyePrincipal principal) {
+    return authorizationManager.generateDatasourceConnection(principal);
```
cyrilou242, line 182: let's have an `authorizationManager.ensureCanRead` call first to ensure the user can read datasources. Else it could leak Pinot tokens.

as of today the design is not that great, so you will have to create a dummy datasource object :/,  see https://github.com/startreedata/thirdeye/blob/d8affb568fd3e8c7977be4208a9be6151591197b/thirdeye-server/src/main/java/ai/startree/thirdeye/service/AnomalyService.java#L100

thirdeye-server/src/main/java/ai/startree/thirdeye/resources/DataSourceResource.java
```diff
@@ -133,6 +133,16 @@ public Response onboardAll(
     return respondOk(onboarded);
   }
 
+  @POST
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
```
cyrilou242, line 137: not necessary I think

cyrilou242 (changes requested):

thirdeye-server/src/main/java/ai/startree/thirdeye/service/DataSourceService.java
anshul98ks123: Yeah, i was thinking to add something like this
but then thought we there will be checks at time of creation
but yeah we need to protect from pinot token data as well

thirdeye-server/src/main/java/ai/startree/thirdeye/service/DataSourceService.java
```diff
@@ -175,4 +178,15 @@ public boolean validate(final ThirdEyePrincipal principal, final long id) {
     authorizationManager.ensureCanRead(principal, dataSourceDto);
     return dataSourceCache.getDataSource(dataSourceDto).validate();
   }
+
+  public DataSourceApi recommend(final ThirdEyePrincipal principal) {
+    final String namespace = authorizationManager.currentNamespace(principal);
+
+    // verify that user has read access to data sources in given namespace
+    DataSourceDTO sampleDataset = new DataSourceDTO();
```
cyrilou242, line 186: final

cyrilou242 (approved):
LGTM!

thirdeye-server/src/main/java/ai/startree/thirdeye/service/DataSourceService.java
anshul98ks123: done

PR #1588, opened 2024-10-09
chore(ui): TE-2472 impact dashboard improvements

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2472

#### Description

Feedback listed in the above tickets has been addressed

### DIFF
--- a/thirdeye-ui/src/app/components/impact-dashboard/summary/summary.component.tsx
+++ b/thirdeye-ui/src/app/components/impact-dashboard/summary/summary.component.tsx
@@ -38,6 +38,7 @@ import { useSummaryData } from "./use-summary-data";
 // Interfaces
 import { SummaryProps } from "./summary.interfaces";
 import { useTranslation } from "react-i18next";
+import { SpecType } from "../../../rest/dto/subscription-group.interfaces";
 
 export const Summary = ({
     alerts,
@@ -50,6 +51,7 @@ export const Summary = ({
     selectedAnalysisPeriod,
     onAnalysisPeriodChange,
     analysisPeriods,
+    activeEnumerationItems,
 }: SummaryProps): JSX.Element => {
     const { t } = useTranslation();
     const summaryRef = useRef<HTMLDivElement>(null);
@@ -63,6 +65,7 @@ export const Summary = ({
         investigations,
         subscriptionGroups,
         mostRecentlyInvestigatedAnomalyAlert,
+        activeEnumerationItems,
     });
 
     const summaryCardsData = [
@@ -181,7 +184,8 @@ export const Summary = ({
                 </div>
                 <div ref={summaryRef}>
                     <div>
-                        In the last <b>{verboseSummaryItems.weeks} weeks</b>,&nbsp;
+                        In the last <b>{verboseSummaryItems.weeks} weeks</b>
+                        ,&nbsp;
                         <b>
                             {summaryData.anomalies.detected.count} anomalies
                             were detected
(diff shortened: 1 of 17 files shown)


thirdeye-ui/src/app/components/impact-dashboard/summary/summary.component.tsx
```diff
@@ -181,7 +184,8 @@ export const Summary = ({
                 </div>
                 <div ref={summaryRef}>
                     <div>
-                        In the last <b>{verboseSummaryItems.weeks} weeks</b>,&nbsp;
+                        In the last <b>{verboseSummaryItems.weeks} weeks</b>
```
esakkiraj, line 187: Are we sure `.weeks` value be > 0 ? I think its better to conditionally handle message for weeks value like 0, 1 and >1 . WYDT ?

thirdeye-ui/src/app/components/impact-dashboard/summary/summary.component.tsx
```diff
@@ -190,8 +194,16 @@ export const Summary = ({
                         than the previous {verboseSummaryItems.weeks} weeks.{" "}
                         {summaryData.anomalies.detected.count > 0 && (
                             <>
-                                Notifications about anomalies were sent via
-                                Slack and Email.
+                                Notifications about anomalies were sent via{" "}
```
esakkiraj, line 197: nit: Spaces and inlining items are little hard to read.

thirdeye-ui/src/app/components/impact-dashboard/summary/use-summary-data.tsx
```diff
@@ -77,8 +78,53 @@ export const useSummaryData = ({
...
+                assoaciatedAlerts?.some((assoaciatedAlert) =>
+                    allAlertIds?.includes(assoaciatedAlert)
+                )
+            ) {
+                allChannels.push(
+                    ...subscriptionGroup.specs.map((spec) => spec.type)
+                );
+            }
+        });
+        if (!isEmpty(allChannels)) {
```
esakkiraj, line 101: nit: better to move out the `allChannels` computation outside of `useEffect` (probably `useMemo`) and only leave the `setVerbose...` part inside effect.

thirdeye-ui/src/app/pages/impact-dashboard-page/data/use-impact-dashboard-api-requests.tsx
```diff
@@ -188,12 +207,22 @@ export const useImpactDashBoardApiRequests = ({
         }
     }, [recentAnomalyInvestigatedId]);
 
+    useEffect(() => {
+        if (alerts) {
+            const activeAlertIds = alerts
+                .filter((alert) => alert.active)
+                .map((alert) => alert.id);
+            getEnumerationItems({ alertIds: activeAlertIds });
```
esakkiraj, line 215: by the name of the function i am thinking if this is needed to be inside `useEffect` or otherwise ? what does this function do which to notify react to re-render ?

thirdeye-ui/src/app/components/impact-dashboard/summary/summary.component.tsx
nalin-patidar: Yes weeks value will always be 4, 13, 21

thirdeye-ui/src/app/components/impact-dashboard/summary/use-summary-data.tsx
nalin-patidar: the computation for channel is not heavy. So I believe useMemo would be on overkill here as with useMemo it will have to keep a reference of the dependencies and old computed value also in the memory and then based on the computation it will trigger the recalculation and then the useEffect will get triggered based on the channel.
Hence kept it in the one block only

thirdeye-ui/src/app/pages/impact-dashboard-page/data/use-impact-dashboard-api-requests.tsx
nalin-patidar: its returned from a custom hook which takes care of the re-render on fetch

esakkiraj (approved):

PR #1597, opened 2024-10-14
chore(ui): TE-2452 improve message when no dataset is present during onboarding

nalin-patidar:
…onboarding

#### Issue(s)

https://startree.atlassian.net/browse/TE-2452

#### Description

When onboarding a dataset from a datasource, show a better worded message when no dataset is present in the selected source.

#### Screenshots

Screenshot 2024-10-14 at 8 32 19 PM

### DIFF
--- a/thirdeye-ui/src/app/pages/welcome-page/create-datasource/onboard-datasets/onboard-datasets-page.component.tsx
+++ b/thirdeye-ui/src/app/pages/welcome-page/create-datasource/onboard-datasets/onboard-datasets-page.component.tsx
@@ -244,64 +244,61 @@ export const WelcomeSelectDatasets: FunctionComponent = () => {
                                                 </Alert>
                                             </Box>
                                         }
-                                        isEmpty={tables?.length === 0}
+                                        isEmpty={isEmpty(tables)}
                                     >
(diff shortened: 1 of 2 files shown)


thirdeye-ui/src/app/pages/welcome-page/create-datasource/onboard-datasets/onboard-datasets-page.component.tsx
```diff
@@ -228,66 +233,75 @@ export const WelcomeSelectDatasets: FunctionComponent = () => {
...
-                                                                        name,
-                                                                    }) => name
-                                                                )
-                                                            );
-                                                    }}
-                                                >
-                                                    {t("label.select-all")}
-                                                </Button>
-                                                <Divider />
+                                        {!isEmpty(tables) && (
```
esakkiraj, line 249: nit: Is this check needed ? From the component `EmptyStateSwitch` name it seems the children will be rendered if only `isEmpty` prop is false ?

nalin-patidar: Ahhh good catch. Removed it. Thanks!

PR #1599, opened 2024-10-17
[server] Convert DataSources Recommend endpoint to GET

anshul98ks123:
#### Issue(s)

#### Description

This endpoint should be a GET method instead of POST
PR fixes that

Safe to change since this API is not being consumed anywhere

#### Testing
Existing UTs + Integration Tests
Swagger page sanity test -> should return null
Screenshot 2024-10-17 at 2 05 47 PM


cyrilou242 (changes requested):
:shipit:

PR #1604, opened 2024-10-18
feat(ui): TE-2448 added namespace config tab in configuration page

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2448

#### Description

Users can configure the notification time format and timezone.


cyrilou242:
not sure if I understand this correctly but this should call 
GET workspace-configuration
PUT workspace-configuration

nalin-patidar:
> not sure if I understand this correctly but this should call GET workspace-configuration PUT workspace-configuration

@cyrilou242 I haven't added the API integration. If we are fine with exposing the format and timezone update via a string with no validation on format, I can go ahead and integrate the API. 
I was under the impression we will wrap this up in one go as part of this ticket([TE-2488](https://startree.atlassian.net/browse/TE-2488)). Let me know how should I proceed.

thirdeye-ui/src/app/rest/dto/workspace.interfaces.ts
```diff
@@ -15,3 +15,18 @@
 export interface Workspace {
     id: string | null;
 }
+
+export interface WorkspaceConfiguration {
```
cyrilou242, line 19: does this mean this will have to be updated anytime the schema is updated in the backend ?

thirdeye-ui/src/app/locale/languages/en-us.json
```diff
@@ -630,6 +632,7 @@
         "weeks-lowercase": "weeks",
         "what-went-wrong-and-where": "What went wrong and where?",
         "whole-alert": "Whole alert",
+        "workspace-config": "workspace config",
```
cyrilou242, line 635: this is user facing right ? 
should we use Settings instead also here ?

cyrilou242 (changes requested):

thirdeye-ui/src/app/locale/languages/en-us.json
nalin-patidar: Yes, let me change it to settings

thirdeye-ui/src/app/rest/dto/workspace.interfaces.ts
nalin-patidar: yes for consistency purpose in the code, but it won't break anything. Since all these are build time checks only

cyrilou242: ok thanks

cyrilou242 (approved):
LGTM thanks!

PR #1612, opened 2024-10-23
[bootstrap] add template configuration to namespace configuration

cyrilou242:
Issue:
https://startree.atlassian.net/browse/TE-2491

Description:
This PR adds support for:
- a default workspace configuration - it is exposed in the server config file
  - `time` configuration is now deprecated. It still has precedence over the workspace configuration for the moment
- TemplateConfiguration in the NamespaceConfiguration. It allows some control on how templates are generated.   
- queryLimit control in TemplateConfiguration. The default query limit of templates can be set as a configuration. This is to be compatible with StarTree Cloud free tier.

Testing:
 - unit tests 
 - manual local test (Cyril)
 
Not part of the PR:
- kubernetes helm chart update
- enterprise query limit logic

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
@@ -13,7 +13,9 @@
  */
 package ai.startree.thirdeye.datalayer.bao;
 
+import static ai.startree.thirdeye.spi.util.SpiUtils.optional;
 import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
 
 import ai.startree.thirdeye.datalayer.dao.NamespaceConfigurationDao;
 import ai.startree.thirdeye.spi.config.TimeConfiguration;
@@ -34,12 +36,15 @@ public class NamespaceConfigurationManagerImpl implements NamespaceConfiguration
 
   private final NamespaceConfigurationDao dao;
   private final TimeConfiguration timeConfiguration;
+  private final NamespaceConfigurationDTO defaultNamespaceConfiguration;
 
   @Inject
   public NamespaceConfigurationManagerImpl(final NamespaceConfigurationDao dao,
-      final TimeConfiguration timeConfiguration) {
+      final TimeConfiguration timeConfiguration,
+      final NamespaceConfigurationDTO defaultNamespaceConfiguration) {
     this.dao = dao;
     this.timeConfiguration = timeConfiguration;
+    this.defaultNamespaceConfiguration = defaultNamespaceConfiguration;
   }
 
   public @NonNull NamespaceConfigurationDTO getNamespaceConfiguration(final String namespace) {
@@ -48,7 +53,7 @@ public class NamespaceConfigurationManagerImpl implements NamespaceConfiguration
 
     // namespace config exists, update if components are empty
     if (existingNamespaceConfig != null) {
-      final boolean updateIsRequired = updateDefaults(existingNamespaceConfig);
+      final boolean updateIsRequired = updateDefaults(existingNamespaceConfig, false);
       if (updateIsRequired) {
         final Long namespaceConfigurationId = save(existingNamespaceConfig);
         // FIXME CYRIL ANSHUL - THIS STATEMENT IS USELESS ? save will throw already ?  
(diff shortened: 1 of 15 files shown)


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
```diff
@@ -22,23 +24,28 @@
...
 import java.util.List;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 @Singleton
 public class NamespaceConfigurationManagerImpl implements NamespaceConfigurationManager {
 
   private final NamespaceConfigurationDao dao;
-  private final TimeConfiguration timeConfiguration;
+  private final @Nullable TimeConfiguration timeConfiguration;
```
anshul98ks123, line 38: why is this nullable?

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/NamespaceConfigurationApi.java
```diff
@@ -18,7 +18,10 @@ public class NamespaceConfigurationApi implements ThirdEyeCrudApi<NamespaceConfi
   private Long id;
   private AuthorizationConfigurationApi auth;
 
-  private TimeConfigurationApi timeConfiguration;
+  private TimeConfigurationApi timeConfiguration = new TimeConfigurationApi();
```
anshul98ks123, line 21: why are we using default values in api object? It should always come from db. right? 
we already have default values in config objects

thirdeye-server/src/main/java/ai/startree/thirdeye/config/ThirdEyeServerConfiguration.java
```diff
@@ -68,8 +73,12 @@ public class ThirdEyeServerConfiguration extends Configuration {
   @JsonProperty("sentry")
   private BackendSentryConfiguration sentryConfiguration = new BackendSentryConfiguration();
 
+  @Deprecated // use defaultWorkspaceConfiguration instead
   @JsonProperty("time")
-  private TimeConfiguration timeConfiguration = new TimeConfiguration();
+  private TimeConfiguration timeConfiguration = null;
+
+  @JsonProperty("defaultWorkspaceConfiguration")
+  private NamespaceConfigurationDTO namespaceConfiguration = new NamespaceConfigurationDTO();
```
anshul98ks123, line 81: shouldn't we have a config file here instead of directly using DTO object?
I know it will introduce one more level of mapping from config file to DTO

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/TimeConfigurationDTO.java
```diff
@@ -13,14 +13,37 @@
...
+import ai.startree.thirdeye.spi.Constants;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.joda.time.DateTimeZone;
 
 public class TimeConfigurationDTO {
 
-  private DateTimeZone timezone;
-  private String dateTimePattern;
-  private long minimumOnboardingStartTime;
+  private static final long JAN_1_2000_UTC = 946684800000L;
```
anshul98ks123, line 24: again, should we have defaults here
or in TimeConfiguration spi config file

anshul98ks123 (changes requested):

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/NamespaceConfigurationManagerImpl.java
cyrilou242: because this is deprecated and will be removed 
in the server config file, it is now valid to not set the `time` config

thirdeye-server/src/main/java/ai/startree/thirdeye/config/ThirdEyeServerConfiguration.java
cyrilou242: yes in theory - it was a hack to get this done quickly.
the config should always be a perfect mapping of the DTO though. 
So not sure what's best. 
Introducing another level results in a lot of useless boilerplate.
But decoupling is still helpful at the configuration level for backward compatibility.

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/NamespaceConfigurationApi.java
cyrilou242: indeed, just the result of me trying things removing this

anshul98ks123 (approved):

thirdeye-server/src/main/java/ai/startree/thirdeye/config/ThirdEyeServerConfiguration.java
cyrilou242: as discussed we will proceed with the current implem and introduce a configuration level once the DTO has stabilized or if we need to decouple DTO and config.

PR #1634, opened 2024-11-05
[pinot] Prefer mytables endpoint over tables endpoint if available

anshul98ks123:
#### Issue(s)

[Prefer mytables endpoint over tables endpoint](https://startree.atlassian.net/browse/TE-2502)

#### Description

Currently, for fetching all the table names during dataset onboarding flow, `/tables` endpoint is called
This doesn't work properly in shared saas setup, where we have workspace aware tokens instead of master token, and those tokens don't have enough permissions to invoke this method due to RBAC policies

whereas, `/mytables` endpoint works in such setup as it manually checks all the individual tables for RBAC authz

This PR is adding fallback logic to always prefer `/mytables` endpoint over `/tables` endpoint

#### Testing

- [X] Manual tests (welcome onboarding flow, dataset onboarding flow)


cyrilou242 (approved):
LGTM!

PR #1637, opened 2024-11-06
[core] fix dataset mapper

cyrilou242:
high priority
this is breaking the enterprise TE cohort recommender
see internal discussion


anshul98ks123 (approved):
LGTM. Thanks

PR #1644, opened 2024-11-08
fix(ui): TE-XX change default algorithm for simple alerts from mean variance to threshold

harshilvelotio:
#### Issue(s)

#### Description

#### Checklist

#### Screenshots

#### How to test

### DIFF
--- a/thirdeye-ui/src/app/components/alert-wizard-v2/alert-template/alert-template.utils.ts
+++ b/thirdeye-ui/src/app/components/alert-wizard-v2/alert-template/alert-template.utils.ts
@@ -159,6 +159,25 @@ export function hasRequiredPropertyValuesSet(
     });
 }
 
+export function createNewStartingAlertThreshold(): EditableAlert {
+    return {
+        name: "",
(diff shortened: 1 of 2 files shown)


cyrilou242 (changes requested):
when selecting the threshold template, the min/max values are defaulted to 0 - 1
the min max values should not be set

thirdeye-ui/src/app/components/alert-wizard-v2/alert-template/alert-template.utils.ts
```diff
@@ -159,6 +159,25 @@ export function hasRequiredPropertyValuesSet(
     });
 }
 
+export function createNewStartingAlertThreshold(): EditableAlert {
+    return {
+        name: "",
```
cyrilou242, line 164: is it ok to set the name to "data-preview" ? 
some errors returned by ThirdEye contain the name, and I think this name would help to understand what happens.

cyrilou242 (changes requested):

PR #1646, opened 2024-11-08
fix(ui): TE-XX show detection props in advance panel

nalin-patidar:
#### Issue(s)

#### Description

#### Checklist

#### Screenshots

#### How to test


thirdeye-ui/src/app/components/alert-wizard-v3/anomalies-filter-panel/anomalies-filter-panel.utils.ts
```diff
@@ -53,6 +53,18 @@ export const getAvailableFilterOptions = (
                     filteringPropertiesBySubStep;
             }
         }
+        if (property.step === "DETECTION") {
+            if (
+                property.name !== "lookback" &&
```
cyrilou242, line 58: is the intent to not show properties that are already shown in the main flow ? 
as implemented this won't work for other templates

cyrilou242 (changes requested):

PR #1647, opened 2024-11-10
fix(ui): TE-XX create alert dx-query-fix

nalin-patidar:
#### Issue(s)

#### Description

#### Checklist

#### Screenshots

#### How to test


cyrilou242:
it's not clear to me how/why the data can be displayed (meaning the template used was correct), but then the template value for the modal comes from somewhere else

nalin-patidar:
> it's not clear to me how/why the data can be displayed (meaning the template used was correct), but then the template value for the modal comes from somewhere else

Its because the the state management is bit scattered. In normal flow the alert has a local state of template, but when you click on Create button, the modal gets it state from a Higher order component. And there the logic for template name with query in its name was miising.

PR #1649, opened 2024-11-11
[core] Add guardrails to prevent same dataset from onboarding twice

anshul98ks123:
#### Issue(s)

[Add guardrails to prevent same dataset(s) from onboarding twice](https://startree.atlassian.net/browse/TE-2437)

#### Description

If for some reason, same datasets are onboarded twice within split second, both are onboarded and it results in an illegal state of DB, and entire functionality breaks -> user is not able to create alerts

This happens in welcome flow, where at dataset onboarding stage, if user accidentally clicks "onboard dataset" button twice before the first operation is completed, datasets are onboarded twice. 

This issue happens only in case of null namespace because in case of non-null namespace, uniqueness constraint in _dataset_config_index_ rejects double persistence of same dataset within a particular namespace ([error log](https://github.com/user-attachments/files/17702226/index_uniqueness_constraint.txt))

This PR adds backend guardrail to check for any existing dataset with given name in same namespace before persisting it to avoid such situations

#### Testing
- [X] Existing UTs
- [X] Existing Integration Tests
- [X] Manual testing

### DIFF
--- a/thirdeye-core/src/main/java/ai/startree/thirdeye/datasource/DataSourceOnboarder.java
+++ b/thirdeye-core/src/main/java/ai/startree/thirdeye/datasource/DataSourceOnboarder.java
@@ -129,6 +129,11 @@ public class DataSourceOnboarder {
     final List<MetricConfigDTO> metrics = datasetConfigDTO.getMetrics();
     datasetConfigDTO.setAuth(auth);
     datasetConfigDTO.setMetrics(null);
+    final DatasetConfigDTO existingDatasetConfig = datasetConfigManager.findUniqueByNameAndNamespace(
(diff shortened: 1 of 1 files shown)


cyrilou242 (approved):
NOTE: 
A common use case is to re-onboard datasets to take into account changes in the schema (eg a new derived column was just added). 

At some point we should make it possible to update the existing dataset configuration in a safe manner, while maintaining custom fields of the dataset. 

At least this change prevents TE to fall in a bad state so let's start with this. See https://startree.atlassian.net/browse/TE-2412

anshul98ks123:
> https://startree.atlassian.net/browse/TE-2412

Thanks for sharing this. We can work on this next whenever we get some bandwidth

PR #1654, opened 2024-11-15
fix(ui): TE-2481 hide need help button

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2481

#### Description

Need help button is hidden till we get it redesigned


cyrilou242 (approved):
nice thanks

PR #1656, opened 2024-11-15
[worker] add pessimistic lock task acquisition

cyrilou242:
Problem: 
The current task acquisition logic uses an optimistic lock. Read  https://en.wikipedia.org/wiki/Optimistic_concurrency_control if necessary. 
Tasks are picked **in order** by every worker.
In order was not enforced before, so there was less chance of contention.
In order is now enforced, so contention is more likely to happen.
At around 5 workers (~4 threads?), workers are very likely to select the same task and spend most of their time rolling back instead of acquiring the task and making progress. 

This PR implements a pessimistic lock, and uses the SKIP LOCKED statement to avoid lock contention.
This should be roughly as fast under small load / small number of workers, and this should scale better with high number of workers (>5).
Realistic benchmarks will be necessary and are not in the scope of this PR.

The feature is disabled by default for the moment. 
It can be enabled by setting `taskDriver.newAcquisitionLogic: true` in the server config.

Tests:
- [X] simulated and checked locks manually for concurrent `SELECT ... FOR UPDATE SKIP LOCKED` queries and `UPDATE`
- [X] end to end tests pass on the new logic 
- [X] unit tests pass on the new logic

NOTE: ⚠️ 
The logic here: https://github.com/startreedata/thirdeye/blob/4951e72be61d4a1b4c0b4665a66c6c722e9fbc44/thirdeye-worker/src/main/java/ai/startree/thirdeye/worker/task/TaskDriverRunnable.java#L221
is slightly incorrect because nextTask could be null because the exception was caught in lower levels. 
This means the sleeping logic is incorrect and the metrics are incorrect. This bug was already present and fixing it requires a pretty big change, it will be done in another PR.

Next steps: 
- [ ] fix bug above 
- [ ] expose `taskDriver.newAcquisitionLogic` in helm chart
- [ ] simulation and benchmarks with high number of workers
  - [ ] test behavior of old and new logic under high number of workers
  - [ ] test impact of lock on other task APIs (should not have any impact)  

References: 
- FOR UPDATE SKIP LOCKED: https://dev.mysql.com/blog-archive/mysql-8-0-1-using-skip-locked-and-nowait-to-handle-hot-rows/
- locking gotchas https://forums.mysql.com/read.php?22,83098,83565#msg-83565
- SKIP LOCKED doc: https://dev.mysql.com/doc/refman/8.4/en/innodb-locking-reads.html
- implementation examples:
  - queue principle https://www.bigbinary.com/blog/solid-queue
  - https://vladmihalcea.com/database-job-queue-skip-locked/
- Investigating locks:
  - https://lefred.be/content/mysql-8-0-data-locking-visibility/
  - https://emmer.dev/blog/investigating-locks-in-mysql/

### DIFF
--- a/thirdeye-integration-tests/src/test/resources/anomalyresolution/config/server.yaml
+++ b/thirdeye-integration-tests/src/test/resources/anomalyresolution/config/server.yaml
@@ -63,6 +63,7 @@ taskDriver:
   noTaskDelay: 1
   randomDelayCap: 1
   taskFailureDelay: 1
+  newAcquisitionLogic: true
 
 scheduler:
   # Run the Quartz Scheduler.
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
@@ -19,15 +19,19 @@ import ai.startree.thirdeye.datalayer.DatabaseClient;
 import ai.startree.thirdeye.datalayer.DatabaseOrm;
 import ai.startree.thirdeye.datalayer.entity.TaskEntity;
 import ai.startree.thirdeye.datalayer.mapper.TaskEntityMapper;
+import ai.startree.thirdeye.datalayer.util.GenericResultSetMapper;
 import ai.startree.thirdeye.spi.Constants;
 import ai.startree.thirdeye.spi.datalayer.DaoFilter;
 import ai.startree.thirdeye.spi.datalayer.Predicate;
 import ai.startree.thirdeye.spi.datalayer.dto.TaskDTO;
+import ai.startree.thirdeye.spi.task.TaskStatus;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.base.Preconditions;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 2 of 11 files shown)


thirdeye-integration-tests/src/test/resources/anomalyresolution/config/server.yaml
```diff
@@ -63,6 +63,7 @@ taskDriver:
   noTaskDelay: 1
   randomDelayCap: 1
   taskFailureDelay: 1
+  newAcquisitionLogic: true
```
anshul98ks123, line 66: Until we have not rolled out the new flow completely
should we keep some integration tests for old logic as well? 
Otherwise any faulty changes to old logic might not get caught

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
```diff
@@ -264,6 +290,42 @@ public List<TaskDTO> executeParameterizedSQL(final String parameterizedSQL,
...
+                return null;
+              } else {
+                throw new RuntimeException(
+                    "Failed to acquire task. Query returned multiple rows. Only one row was expected. Please reach out to StarTree support.");
+              }
+            }
+          }, null);
+    } catch (SQLException e) {
+      LOG.error(e.getMessage(), e);
+      return null;
```
anshul98ks123, line 325: shouldn't we raise an exception here?
maybe that's why waitForTask() is not receiving any errors in exception block

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
```diff
@@ -19,15 +19,19 @@
...
 import ai.startree.thirdeye.spi.datalayer.Predicate;
 import ai.startree.thirdeye.spi.datalayer.dto.TaskDTO;
+import ai.startree.thirdeye.spi.task.TaskStatus;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.base.Preconditions;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
```
cyrilou242, line 34: test

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
cyrilou242: yes we may want to throw
the real confusion comes from line 322. 
databaseClient.executeTransaction already catches all issues, except issues that happen when trying to get a Connection object.  

--> so in effect it's almost impossible to get into this catch clause.

I will do this fix in another PR. I started it but realized it was too big and impacted many other places so I removed it from this PR.
IMO it's fine if the databaseClient.executeTransaction throws exception for any error, and then it really depends on the consumer whether they want to catch the exception or not. 
The fact that databaseClient.executeTransaction does not even catch all exceptions means the callers still have to add a try catch, which results in duplication of default values and very messy code.

thirdeye-integration-tests/src/test/resources/anomalyresolution/config/server.yaml
cyrilou242: yeah I thought about this. 
the legacy logic really should not be changed anymore, and I'd like to push for the new one aggressively.
The codepaths reached by the old one are pretty obvious and it will be easy to catch if someone (= you or me) edits it. If it's the case we can always run on the old logic by changing this value manually.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
```diff
@@ -264,6 +290,42 @@ public List<TaskDTO> executeParameterizedSQL(final String parameterizedSQL,
...
+    try {
+      return databaseClient.executeTransaction(
+          connection -> {
+            try (final Statement s = connection.createStatement();
+                final ResultSet rs = s.executeQuery(SELECT_AND_LOCK_NEXT_TASK_QUERY)
+            ) {
+              final List<TaskEntity> res = genericResultSetMapper.mapAll(rs, TaskEntity.class);
+              if (res.size() == 1) {
+                final List<TaskDTO> dtos = toDto(res);
+                final TaskDTO toUpdate = dtos.get(0);
```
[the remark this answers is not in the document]
cyrilou242: this PR was created before java 21 got merged :) 
it's still running on java 17 here and I won't rebase just for this, I'll do the change in the next PR that fixes the databaseClient interface.

anshul98ks123 (approved):

PR #1658, opened 2024-11-18
[thirdeye-pinot] Enrich pinot controller schema fetch error log

anshul98ks123:
Error log when fetching schema from pinot controller currently doesn't log exception object
This PR adds exception obj to the log


cyrilou242 (approved):
:shipit:

PR #1665, opened 2024-11-19
[server] Improve related entities logic in AuthZ Manager

anshul98ks123:
#### Issue(s)

[impossible to delete alerts](https://startree.atlassian.net/browse/TE-2554)

#### Description
Due to related entities check in authorization manager layer, if a dataset is deleted which already has an alert setup on it, every alert related request fails.

This leads to a state where any screen which has to load alerts in some form breaks (even the home page)
and it's impossible to delete the broken alert, either from UI or from swagger APIs

If a related entity has been deleted, then it's not an ideal state but we shouldn't break entire call
This PR fixes the related entities logic by relaxing it in such cases and logs error for further debugging

#### Testing

- [X] Existing UTs
- [X] Existing Integration Tests
- [X] Manual Testing

1. Run local server
2. Delete a dataset, which is already associated to an alert
3. Check home page
4. Check alert page of broken alert

Before:
https://github.com/user-attachments/assets/9814e1aa-4064-4030-8efa-de54adbbd170

After:
https://github.com/user-attachments/assets/1cf8df20-fb07-4510-b464-089bc4048448

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/auth/AuthorizationManager.java
@@ -372,6 +372,7 @@ public class AuthorizationManager {
   private <T extends AbstractDTO> void addRelatedEntities(final T entity,
       final Set<ResourceIdentifier> result,
       final Caches caches) {
+
     if (entity instanceof final AlertDTO alertDto) {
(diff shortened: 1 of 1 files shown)


cyrilou242 (approved):
LGTM thanks!

cyrilou242:
merging to cut a release with this fix

PR #1696, opened 2024-11-28
[server] make only the scheduler register expensive database metrics

cyrilou242:
Some metrics are based on queries run on the MySql instance. 
As of today they are registered when DAO singletons are instantiated. 
This means they are registered for all components of ThirdEye. (coordinator, scheduler, worker).
This causes two issues: 
- it is confusing
  - because there is a cache, the value of the metric may not match between components. The difference can be huge.
  - when writing prometheus queries, one must be wary of not aggregating the same thing twice
- when scaling workers or coordinators, the number of metrics queries scale linearly. It should not.

This PR implements logic to only have the scheduler collect the expensive database-based metrics.

A log is added to make it simple to check if the metrics are registered correctly:
```
INFO  [2024-11-28 16:15:29,616] ai.startree.thirdeye.datalayer.bao.AnomalyManagerImpl: Registered anomaly database metrics.
INFO  [2024-11-28 16:15:29,616] ai.startree.thirdeye.datalayer.bao.TaskManagerImpl: Registered task database metrics.
INFO  [2024-11-28 16:15:29,617] ai.startree.thirdeye.datalayer.bao.RcaInvestigationManagerImpl: Registered RCA investigation database metrics.
INFO  [2024-11-28 16:15:29,617] ai.startree.thirdeye.datalayer.bao.AlertManagerImpl: Registered alert database metrics.
INFO  [2024-11-28 16:15:29,617] ai.startree.thirdeye.datalayer.bao.SubscriptionGroupManagerImpl: Registered subscription group database metrics.
```

these logs should only happen once in the lifecycle of a scheduler, at the start.

Main changes are in `ThirdEyeServer` and `AbstractManager`.
ThirdEyeServer finds instances of AbstractManager and calls registerDatabaseMetrics.

### DIFF
--- a/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/persistence/DelegateDatasetConfigManager.java
+++ b/thirdeye-detectionpipeline/src/main/java/ai/startree/thirdeye/detectionpipeline/persistence/DelegateDatasetConfigManager.java
@@ -132,4 +132,9 @@ public class DelegateDatasetConfigManager implements DatasetConfigManager {
   public List<DatasetConfigDTO> findActive() {
     return delegate.findActive();
   }
+
+  @Override
+  public void registerDatabaseMetrics() {
+    delegate.registerDatabaseMetrics();
+  }
 }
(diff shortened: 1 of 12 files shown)


anshul98ks123 (approved):
:shipit:

PR #1698, opened 2024-12-01
fix(ui): TE-2216 bug fixes for simple alert

harshilvelotio:
##### Issue 1
```
in dx --> sql query, configure notifications appear above the charts instead of below
```
##### Issue 2
```
let's remove the title here
this may mean the "json editor" will be moved to the row below. (not sure)
goal is to gain some vertical space 
```
##### Issue 3
```
"json editor" is not a json editor anymore
let's call it "VIEW JSON" for the moment
it may be ok to put it next to "add advanced options" for the moment
let's make sure the behavior is correct when the template is not selected yet. The button can be disabled, or it can show a modal with the config that was used for data preview, whatever is the simplest. 
```
##### Issue 4
```
in the JSON modal, remove cancel and apply changes. Just put a CLOSE button
```

##### Issue 5
```
I seem to remember we discussed having a dimension search bar in the dx case.
```

#### Description

#### Checklist

#### Screenshots

#### How to test

### DIFF
--- a/thirdeye-ui/src/app/components/additional-filters-drawer/additional-filters-drawer.component.tsx
+++ b/thirdeye-ui/src/app/components/additional-filters-drawer/additional-filters-drawer.component.tsx
@@ -14,7 +14,16 @@
  */
 
 import { Icon } from "@iconify/react";
-import { Box, Button, Drawer, Typography } from "@material-ui/core";
+import {
+    Accordion,
+    AccordionDetails,
+    AccordionSummary,
+    Box,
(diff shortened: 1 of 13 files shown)


cyrilou242:
Screenshot 2024-12-04 at 10 32 09

--> search dimensions at the bottom right corner

cyrilou242:
Screenshot 2024-12-04 at 10 34 06
--> this looks bad, please create a dedicated PR for this

PR #1707, opened 2024-12-05
[core] fix datasource metric gets broken in case of incorrect datasource configuration

cyrilou242:
getDataSource can throw exceptions
with the previous implementation, any exception was breaking the metric. 

change: catch the exception


anshul98ks123 (approved):
:shipit:

PR #1709, opened 2024-12-06
[persistence] remove namespace migration logic that could cause oom in edge cases

cyrilou242:
update temporary code paths that could result in OOM when the number of anomalies for many alerts is high and multiple stats call are performed concurrentl (this can happen in the landing page)


anshul98ks123 (approved):

anshul98ks123:
:shipit:

PR #1712, opened 2024-12-09
fix(ui): Added Accordion to the input groups in drawer and increase the drawer width

harshilvelotio:
#### Issue(s)

#### Description

Added Accordion to the input groups in drawer and increase the drawer widthDescribe the changes being introduced to double for better view . 

#### Checklist

#### Screenshots

Screenshot 2024-12-09 at 6 00 23 PM

#### How to test


cyrilou242:
we should have all template properties, 
(except the ones already present in the main flow of course (eg min, max))
accordions should be on the section level.
time of week, guardrail metric, special events are sub sections

cyrilou242:
the order of sections should always be the same and match the order in the advanced mode

Screenshot 2024-12-16 at 09 12 24
Screenshot 2024-12-16 at 09 12 51
Screenshot 2024-12-16 at 09 13 10

PR #1713, opened 2024-12-09
[persistence] remove exception catch in DatabaseClient

cyrilou242:
See https://startree.atlassian.net/browse/TE-2579

Fix DatabaseClient exception throwing design. DatabaseClient does not catch exception anymore, it's the consumer that is responsible for catching the error. 
From what I saw having the DAOs catching error is still not a good design because they log the errorn, replace with an error values (0, null), and then the consumer still has to perform an argument/state check on the error value. If an error value is encountered, the an exception is thrown but it does not contain the stack of the original error anymore.
This can be taken in another PR. 

Has the effect of fixing the issue of `TaskManager(Impl).acquireNextTaskToRun` not surfacing errors because they were caught by the DatabaseClient, ie it fixes a task scheduler bug.
https://github.com/startreedata/thirdeye/pull/1713/files#diff-d93736e73d043688707a5ad2c487338a9ec51dcf0bed01f6a3d55e7783d17330L216
This bug needs to be fixed for proper benchmarking.

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/DatabaseClient.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/DatabaseClient.java
@@ -52,8 +52,8 @@ public class DatabaseClient {
         .register(Metrics.globalRegistry);
   }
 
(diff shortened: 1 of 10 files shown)


anshul98ks123 (approved):
:shipit:

PR #1714, opened 2024-12-10
[server] Handle PinotClientException in exception handler

anshul98ks123:
#### Issue(s)

[Surface the pinot quota error properly](https://startree.atlassian.net/browse/TE-2580)

#### Description

Currently, there are several problems in how pinot exceptions are caught and surfaced.
1. Errors thrown by PinotQueryExecutor's [load()](https://github.com/startreedata/thirdeye/blob/302335a30554a628eb1ce5f20d63991d8e71ae07/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotQueryExecutor.java#L216) method are not caught properly in the [caller code](https://github.com/startreedata/thirdeye/blob/302335a30554a628eb1ce5f20d63991d8e71ae07/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotThirdEyeDataSource.java#L161) because any exception thrown in load method of a guava cache is wrapped around by UncheckedExecutionException, which is not being caught here
2. These exceptions are further down the line wrapped around by ExecutionException but they're not being handled properly in [ExceptionHandler for RCA case](https://github.com/startreedata/thirdeye/blob/302335a30554a628eb1ce5f20d63991d8e71ae07/thirdeye-server/src/main/java/ai/startree/thirdeye/core/ExceptionHandler.java#L79)
3. Exceptions thrown by PinotQueryExecutor emit only the sql command in their error message, not the root exception reason.

This results into final exception to be thrown of the form:
```
{
    "code": "ERR_UNKNOWN_RCA_ALGORITHM",
    "msg": "Unknown error running the rca algorithm: java.util.concurrent.ExecutionException: org.apache.pinot.client.PinotClientException: Error when running SQL: ...",
}
```

This PR:
1. Fixes ExceptionHandler for RCA case to catch ExecutionExceptions and further checks if underlying exception is a ThirdEyeException (which is thrown in pinot client). If yes, then it throws that exception with same cause and message, else use a custom cause.
2. Handles UncheckedExecutionException exceptions when PinotQueryExecutor cache get() (underlying load()) is called.
4. Currently, there's no proper way to know error code in resultSetGroup returned by the Pinot client, hence for short term workaround, we have added a substring check for QuotaExceededError error.

Post these fixes, pinot client exceptions are caught and surfaced correctly in following form:
```
{
    "code": "ERR_PINOT_QUERY_EXECUTION",
    "msg": "Pinot Query Execution Failed. Error: org.apache.pinot.client.PinotClientException: Query had processing exceptions: \n[{\"message\":\"null:\\n1 segments unavailable: [pageviews_with_nulls_OFFLINE_20200202_20200730_0]\",\"errorCode\":305}]. SQL: ...",
}
```

#### Testing

- [X] Existing UTs
- [X] Existing Integration tests
- [X] Manual test of `/api/rca/metrics/heatmap` API via swagger

| Heatmap Response Before  | Heatmap Response After |
| ------------- | ------------- |
| [rca-heatmap-before.json](https://github.com/user-attachments/files/18094664/rca-heatmap-before.json)  | [rca-heatmap-after.json](https://github.com/user-attachments/files/18094666/rca-heatmap-after.json) |

### DIFF
--- a/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotQueryExecutor.java
+++ b/thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/PinotQueryExecutor.java
@@ -13,10 +13,14 @@
  */
 package ai.startree.thirdeye.plugins.datasource.pinot;
 
+import static ai.startree.thirdeye.spi.ThirdEyeStatus.ERR_PINOT_QUERY_EXECUTION;
+import static ai.startree.thirdeye.spi.ThirdEyeStatus.ERR_PINOT_QUERY_QUOTA_EXCEEDED;
+
 import ai.startree.thirdeye.plugins.datasource.pinot.resultset.ThirdEyeDataFrameResultSet;
 import ai.startree.thirdeye.plugins.datasource.pinot.resultset.ThirdEyeResultSet;
 import ai.startree.thirdeye.plugins.datasource.pinot.resultset.ThirdEyeResultSetGroup;
 import ai.startree.thirdeye.plugins.datasource.pinot.resultset.ThirdEyeResultSetMetaData;
+import ai.startree.thirdeye.spi.ThirdEyeException;
 import ai.startree.thirdeye.spi.dataframe.DataFrame;
 import ai.startree.thirdeye.spi.detection.v2.ColumnType;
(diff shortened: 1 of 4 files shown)


cyrilou242 (approved):
LGTM thanks!

PR #1716, opened 2024-12-10
[persistence] optimize db metrics caching by scheduling and spreading refresh load

cyrilou242:
some metrics are expensive to compute
with `memoizeWithExpiration`, the computation is performed when the prometheus endpoint is hit.
if there are many metrics supplier that expired at the same time (which is very likely), the computation is triggered for all these metrics at the same time. This is stressing the db and also result in prometheus parse timeout.  

This PR introduce a helper to schedule the refresh of a metric instead of waiting for the prometheus hit. Some random delay is added to spread the request load (the spread may not be optimal but it's ok to begin with)  

Let me know what you think.


anshul98ks123 (approved):
LGTM. Thanks

PR #1719, opened 2024-12-13
te 2587 optimize task metrics queries

cyrilou242:
- **clean**
- **clean**
- **[persistence] optimize task count operations and metrics**

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
@@ -372,7 +372,9 @@ public class TaskManagerImpl implements TaskManager {
 
   @Override
   public List<TaskDTO> findAll() {
-    return dao.getAll();
+    // this operation can stress the task execution framework - if there is no use case for it, keep it unsupported
+    // most use cases should use a filter by namespace
+    throw new UnsupportedOperationException("findAll operation is not supported for Tasks. If this error is unexpected, please reach out to support.");
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


anshul98ks123 (approved):

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -205,12 +207,15 @@ public void purge(@Nullable final Duration expiryDurationOptional,
 
     final long startTime = System.nanoTime();
     final List<TaskDTO> tasksToBeDeleted = filter(new DaoFilter()
+        // non locking read
+        .setTransactionIsolationLevel(Connection.TRANSACTION_READ_UNCOMMITTED)
```
anshul98ks123, line 211: tasks purge is a core operation and not some metric collection. Right?

Are we doing this to speed up this process 
because even if some tasks are uncommitted, they'll be purged in next cycle properly anyways?

cyrilou242: this query is doing a full scan (createTime index will be used but logically it's still a full scan), so it would prevent inserts and updates on the table 
so it blocks the 
this means the 

this query is still pretty bad because it loads 5000 tasks in memory, just to delete them. 
but I'll fix this later.

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/TaskDao.java
```diff
@@ -231,18 +217,27 @@ public List<TaskDTO> filter(final DaoFilter daoFilter) {
...
 
   public List<TaskDTO> get(final Predicate predicate) {
     return get(predicate, null);
   }
   
   public List<TaskDTO> get(final Predicate predicate, @Nullable Long limit) {
+    return get(predicate, limit, null);
+  }
+
+  public List<TaskDTO> get(final Predicate predicate, @Nullable Long limit, @Nullable Integer transactionIsolationLevel) {
```
cyrilou242, line 231: final

PR #1722, opened 2024-12-13
[persistence] optimize first heartbeat

cyrilou242:
See https://startree.atlassian.net/browse/TE-2586

We can skip the first heartbeat by setting the `lastActive` value at acquisition directly. 
this saves 2 queries for all task that run faster than the heartbeat frequency. 

For most detection, this means the number of queries per task is reduced by 2.
At high scale this matters. The heartbeat queries take locks.

Note:
the heartbeat logic is still bad because it has a concurrency issue but this will be taken in another PR. See https://startree.atlassian.net/browse/TE-2578

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
@@ -129,6 +129,7 @@ public class TaskManagerImpl implements TaskManager {
     return dao.acquireNextTaskToRun(workerId);
   }
 
+  @Deprecated
   @Override
   public TaskDTO findNextTaskToRun() {
     final String queryClause = """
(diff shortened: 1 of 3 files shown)


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
```diff
@@ -147,13 +148,15 @@ AND ref_id not in (select ref_id from task_entity where status = 'RUNNING')
    * This method has side effects on the task DTO, even if the acquisition attempt fails.
    * Re-fetch the taskDto if you need to ensure consistency with the persistence layer.
    */
+  @Deprecated
```
anshul98ks123, line 151: why are we deprecating this method?

cyrilou242: it was already deprecated in the interface, it's only used by the legacy task acquisition logic
same for the other one

anshul98ks123 (approved):
:shipit:

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/TaskManagerImpl.java
cyrilou242: `acquireNextTaskToRun`
should be used instead, it performs find and acquire in a single optimized way

PR #1725, opened 2024-12-16
Add Jacoco Code Coverage support

anshul98ks123:
### Issue(s)

[[BE] Apps Team Testing & Secure CI Update - V1 (Manual)](https://startree.atlassian.net/browse/AI-18)

### Description

This PR adds basic jacoco code coverage local setup

### How to test
1. `./mvnw clean verify`
2. Check aggregate coverage report in 
```jacoco-code-coverage/target/site/jacoco-aggregate/index.html```

Screenshot 2024-12-16 at 8 57 03 PM

Also verified that maven release doesn't break by a local run
```
./mvnw clean -B -DskipTests -Darguments=-DskipTests release:clean initialize release:prepare -U
```
Output:
[thirdeye-release-logs.txt](https://github.com/user-attachments/files/18151998/thirdeye-release-logs.txt)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
     <module>pinot-test-container</module>
     <module>thirdeye-integration-tests</module>
     <module>thirdeye-benchmarks</module>
+    <module>jacoco-code-coverage</module>
   </modules>
 
   <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -71,8 +71,6 @@
     <hsqldb.version>2.7.1</hsqldb.version>
     <!-- should match dropwizard version - except for security issues -->
     <jackson.version>2.18.1</jackson.version>
-    <!-- Skip Jacoco by default -->
-    <jacoco.skip>true</jacoco.skip>
     <jodatime.version>2.12.7</jodatime.version>
     <maven-assembly-plugin.version>3.3.0</maven-assembly-plugin.version>
     <mockito.version>5.12.0</mockito.version>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pom.xml [hunk 1]
```diff
@@ -70,6 +71,8 @@
     <hsqldb.version>2.7.1</hsqldb.version>
     <!-- should match dropwizard version - except for security issues -->
     <jackson.version>2.18.1</jackson.version>
+    <!-- Skip Jacoco by default -->
+    <jacoco.skip>true</jacoco.skip>
```
cyrilou242, line 75: what's the test time difference between jacoco enabled and disabled ? 
if it's not too much maybe we could have it run in all PRs

anshul98ks123: just 3 second difference
lemme remove the flag

cyrilou242:
> Also added a profile skip-integration-tests to speedup jacoco report generation, as integration tests module are not needed while computing code coverage and -DskipTests is not useful since it skips all tests

I'm a bit confused by this. Best case jacoco should be able to measure codepaths hit by the integration tests.

anshul98ks123:
> > Also added a profile skip-integration-tests to speedup jacoco report generation, as integration tests module are not needed while computing code coverage and -DskipTests is not useful since it skips all tests
> 
> I'm a bit confused by this. Best case jacoco should be able to measure codepaths hit by the integration tests.

you're right
i had to exclude integration tests path in dm code because it was taking like hours of time, but forgot the basic point that integration tests do test codepaths of other modules and contribute to the code coverage score

reverting that commit

anshul98ks123:
done

cyrilou242 (approved):
LGTM
:shipit:

PR #1730, opened 2024-12-17
[scheduler] demo implem of generic task cron scheduler

cyrilou242:
proposal to avoid code duplication between DetectionCronScheduler and SubscriptionGroupScheduler. 

the first commit is a valid implementation that make use of interfaces. 
It's a bit intrusive to existing stable DTOs, but keeps the generic class implementation relatively simple.

The second commit is an implementation that makes use of lambdas. I think it's more future proof and keeps the change isolated to the generic class. 
It makes the generic class more complex.

Overall I'm not sure yet if this is a good change. 
It will be good for performance improvements and reliability, because it will ensure the scheduling of both detection and notifications follows the exact same logic.
It may be a hindrance if we ever need to change the logic in complex ways or need to introduce new tasks that don't need the same logic. 
Also the Generic Runnable has a pretty unnatural constructor with many arguments. (some could be removed with more refactoring out of the scope of this PR)


thirdeye-scheduler/src/main/java/ai/startree/thirdeye/scheduler/TaskCronSchedulerRunnable.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+    this.entityDao = entityDao;
+    this.entityName = entityClazz.getSimpleName();
+    this.taskType = taskType;
+    this.jobClazz = jobClazz;
+    this.cronMaxTriggersPerMinute = cronMaxTriggersPerMinute;
+    this.log = LoggerFactory.getLogger(loggerClass);
+
+    this.groupMatcher = GroupMatcher.jobGroupEquals(taskType.toString());
+
```
anshul98ks123, line 84: nit: remove newline

anshul98ks123:
Overall the code looks good and does indeed prevents exactly same code from being duplicated

Only concern is if these implementations start to diverge or require different handling, how big the lift would be at that time to accommodate that, for a bug or a change

maybe some overriding logic similar to what we have in ThirdEyeAuthorizerProvider that can override the generic implementation, if needed? 🤔

thirdeye-scheduler/src/main/java/ai/startree/thirdeye/scheduler/TaskCronSchedulerRunnable.java
```diff
@@ -0,0 +1,191 @@
...
+      final Class<? extends Job> jobClazz,
+      final GuiceJobFactory guiceJobFactory,
+      final int cronMaxTriggersPerMinute,
+      final Class<?> loggerClass) {
+    try {
+      scheduler = StdSchedulerFactory.getDefaultScheduler();
+      scheduler.setJobFactory(guiceJobFactory);
+      scheduler.start();
+    } catch (final SchedulerException e) {
+      throw new RuntimeException("Failed to initialize the scheduler", e);
```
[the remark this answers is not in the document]
cyrilou242: yes, fixed

thirdeye-scheduler/src/main/java/ai/startree/thirdeye/scheduler/TaskCronSchedulerRunnable.java
cyrilou242: yes thanks

cyrilou242:
> Overall the code looks good and does indeed prevents exactly same code from being duplicated
> 
> Only concern is if these implementations start to diverge or require different handling, how big the lift would be at that time to accommodate that, for a bug or a change
> 

for a bug change under pressure for only one task type, the simplest method would be to copy `TaskCronSchedulerRunnable` eg `TaskCronSchedulerRunnableForSubscription`, and edit whatever needs to be edited. 
This would result in code duplication roughly equivalent to the current state. 
Then the code can be cleaned up to be put back inside the `xxxCronScheduler` if need be. 

> maybe some overriding logic similar to what we have in ThirdEyeAuthorizerProvider that can override the generic implementation, if needed? 🤔
given we don't know where and what will need to be overridden, I don't want to introduce early abstractions, but this makes sense

PR #1738, opened 2025-01-07
feat(ui): TE-2610 show workspace switcher if workspace count is more than 0

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2610

#### Description

We were only showing workspace switcher if the there were more than one workspace. With this change we are showing the switcher if the workspace count is more than 0.


sid-sawhney:
1. When a user has only one workspace, will it show Default(not set) or will it show the workspace id? I'd like the drop down to still show the workspace id or environment name
2. Can we reduce the font size of the workspace name in the drop down to be the same size or smaller than the 'Search Workspace' text in the search bar

PR #1741, opened 2025-01-08
[pinot] Add logic to override maxQPS quota of datasets in onboarding flow

anshul98ks123:
#### Issue(s)

[upgrade qps quota in Pinot table when table is onboarded in TE](https://startree.atlassian.net/browse/TE-2505)

#### Description

In Free Tier, each table created in Data Manager is created with a default table query quota of 5 queries per second. This default table query quota causes many problems for Free Tier ThirdEye as the UI breaches this limit on many surfaces

This PR adds changes to be able to upgrade qps quota when users onboard new tables to ThirdEye to update the configuration of the onboarded table, changing the 5 QPS limit to some higher configurable value (50 / 100 QPS)

One can set pinotMaxQPSQuotaOverride in quotas in server config to configure this feature

#### Testing

- [X] Existing UTs
- [X] Existing Integration Tests
- [X] Manual testing

PR #1746, opened 2025-01-09
[k8s] Add quotas to helm template

anshul98ks123:
#### Issue(s)

[upgrade qps quota in Pinot table when table is onboarded in TE](https://startree.atlassian.net/browse/TE-2505)

#### Description
Adding quotas to helm template configuration
It is used to enable the feature to override the default pinot max QPS quota for the tables that are being onboarded to TE (#1741)

#### How to test
Validate helm install locally
`cd kubernetes/helm/startree-thirdeye`

`helm install --generate-name . --debug --dry-run`
[helm-debug-output-log-plain.txt](https://github.com/user-attachments/files/18361263/helm-debug-output-log-plain.txt)

`helm install --generate-name . --debug --dry-run --set quotas.pinotMaxQPSQuotaOverride=100`
[Uploading helm-debug-output-log-override.txt…]()

### DIFF
--- a/kubernetes/helm/startree-thirdeye/templates/coordinator/configmap.yaml
+++ b/kubernetes/helm/startree-thirdeye/templates/coordinator/configmap.yaml
@@ -135,6 +135,9 @@ data:
       tags:
         component: "{{ .Values.coordinator.name }}"  
 
+    quotas:
+{{ toYaml .Values.quotas | indent 6 }}
+
     notifications:
       useSendgridEmail: true
       smtp:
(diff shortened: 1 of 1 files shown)


cyrilou242 (changes requested):
pretty sure that generating 

```
quotas:
```

will not be a valid config for the server

if think you'll have to wrap 
```
quotas 
```
with a `if`

anshul98ks123:
> pretty sure that generating
> 
> ```
> quotas:
> ```
> 
> will not be a valid config for the server
> 
> if think you'll have to wrap
> 
> ```
> quotas 
> ```
> 
> with a `if`

anshul98ks123:
thanks. addressed

PR #1747, opened 2025-01-11
[server] add demo dataset creation support

cyrilou242:
Add new endpoints to create demo datasets: 
- `DemoDatasetApi`: contains metadata about the demo dataset. Exposed via `GET /{datasource-id}/demo-datasets`. Frontend can use this endpoint to get a list of DemoDatasetApi for a given datasource
- `POST /{datasource-id}/demo-datasets` + `+ a demo dataset id` creates the corresponding demo dataset in the DBMS and onboards the table in ThirdEye.

Demo datasets are per datasource implementation, because some datasource implementation may not (be able to) provide the same demo dataset.

## TODO: 
- [X] add happyPath test
- [X] tested locally 

## Design note: 
1. The table onboarding is slow, see
https://startreedata.slack.com/archives/C019DPR16JW/p1736590160817849?thread_ts=1736537370.189139&cid=C019DPR16JW
This will be picked later, best would be to contribute to Pinot I think.
2. If there is a name conflict (eg a table with the same name already exists), the logic will fail. I think it's ok. It's not clear yet whether: 
  - we should skip the table creation and onboard
  OR
  - we should delete the existing table with the same name
  OR
  - we should create the demo table with a different_name

## Commits 
- **[spi] add demo datasets api**
- **[server] add demo datasets endpoints**
- **add nullable info to api**
- **implement service**
- **implement methods in MeteredDataSource**
- **implement availableDemoDatasets in pinot datasource**
- **clean**
- **implement createDemoDataset in Pinot datasource**


thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/restclient/PinotControllerRestClient.java
```diff
@@ -222,17 +208,87 @@ public void updateTableMaxQPSQuota(final String dataset, final JsonNode tableJso
...
+      HttpUtils.safeClose(response);
     }
 
     return ;
   }
 
+  // schema name should always be equal to table name without the type
+  public @NonNull String postSchema(final Schema schema, final boolean override,
+      final boolean force) throws IOException {
+    final HttpPost request = new HttpPost(
```
anshul98ks123, line 220: nit: const for URL template

thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/restclient/PinotControllerRestClient.java
```diff
@@ -222,17 +208,87 @@ public void updateTableMaxQPSQuota(final String dataset, final JsonNode tableJso
...
+      throw e;
+    } finally {
+      HttpUtils.safeClose(response);
+    }
+    return schema.getSchemaName();
+  }
+
+  // returns the table name with type. Eg pageviews_OFFLINE
+  public @NonNull String postTable(final TableConfig tableConfig) throws IOException {
+    final HttpPost request = new HttpPost("/tables");
```
anshul98ks123, line 244: replace with already existing const PINOT_TABLES_ENDPOINT

thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/restclient/PinotControllerRestClient.java
cyrilou242: imho templated string in constants makes thing less readable. 

eg with
```
SCHEMA_POST_ROUTE.formatted(override, force)
```
it's actually hard to know what happens.

(theoretically it's bad for reading because it's an _indirection_). 

also we only use this once

thirdeye-plugins/thirdeye-pinot/src/main/java/ai/startree/thirdeye/plugins/datasource/pinot/restclient/PinotControllerRestClient.java
```diff
@@ -222,17 +208,87 @@ public void updateTableMaxQPSQuota(final String dataset, final JsonNode tableJso
...
+  }
+  
+  // note: this is slow because this is downloading the file before sending it to Pinot
+  // see discussion here https://startreedata.slack.com/archives/C019DPR16JW/p1736590160817849?thread_ts=1736537370.189139&cid=C019DPR16JW
+  // don't want to introduce a caching on disk strategy, I think the issue should be fixed on Pinot side
+  public void postIngestFromFile(final @NonNull String tableNameWithType,
+      final @NonNull String batchConfigMapStr, final @NonNull String sourceURIStr)
+      throws IOException {
+    final HttpPost request = new HttpPost(
+        "/ingestFromFile?tableNameWithType=%s&batchConfigMapStr=%s"
```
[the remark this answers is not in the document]
cyrilou242: see comment above

PR #1751, opened 2025-01-15
[persistence] Add TaskQuotasConfiguration to workspace configuration

anshul98ks123:
#### Issue(s)

[thirdeye tasks limit per month](https://startree.atlassian.net/browse/TE-2603)

#### Description

[ThirdEye Free Tier Usage Quota
](https://docs.google.com/document/d/10UK1rLjSnwoe-gIKy9K7erOM_1DpMaNOHsGn9jgQ4AM/edit?tab=t.0#heading=h.2k2bjxgh7f56)

This PR introduces TaskQuotasConfiguration as part of workspace configuration
It will be a non-editable field and will be used for limiting the number of detection and notification tasks that can be run in month for a particular workspace in shared saas setup

To configure this for a cluster, one can provide default limits as part of defaultWorkspaceConfiguration in server.yaml 
```
defaultWorkspaceConfiguration:
  timeConfiguration:
    # notification timezone
    timezone: UTC
    # notification time format pattern - see https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html 
    dateTimePattern: "MMM dd, yyyy HH:mm"
    # Limit onboarding alert replay back to JAN_1_2000_UTC
    minimumOnboardingStartTime: 946684800000
  templateConfiguration:
    # alert templates generate queries with a LIMIT statement - this is the default value of the limit statement
    sqlLimitStatement: 100_000_001
  quotasConfiguration:
    taskQuotasConfiguration:
      maximumDetectionTasksPerMonth: 3100
      maximumNotificationTasksPerMonth: 21000
```

#### Testing

- [X] UTs
- [X] Integration Tests
- [X] Manual testing

### DIFF
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/service/NamespaceConfigurationService.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/service/NamespaceConfigurationService.java
@@ -83,10 +83,7 @@ public class NamespaceConfigurationService {
       NamespaceConfigurationApi updated) {
     if (!Objects.equals(existing.getId(), updated.getId()) ||
         !Objects.equals(existing.getAuth().getNamespace(), updated.getAuth().getNamespace()) ||
-        !Objects.equals(existing.getQuotasConfiguration().getTaskQuotasConfiguration().getMaximumDetectionTasksPerMonth(),
-            updated.getQuotasConfiguration().getTaskQuotasConfiguration().getMaximumNotificationTasksPerMonth()) ||
-        !Objects.equals(existing.getQuotasConfiguration().getTaskQuotasConfiguration().getMaximumNotificationTasksPerMonth(),
-            updated.getQuotasConfiguration().getTaskQuotasConfiguration().getMaximumNotificationTasksPerMonth())) {
+        !Objects.equals(existing.getQuotasConfiguration(), updated.getQuotasConfiguration())) {
       throw badRequest(
           ThirdEyeStatus.ERR_NAMESPACE_CONFIGURATION_VALIDATION_FAILED,
           existing.namespace(),
--- a/thirdeye-server/src/main/java/ai/startree/thirdeye/service/NamespaceConfigurationService.java
+++ b/thirdeye-server/src/main/java/ai/startree/thirdeye/service/NamespaceConfigurationService.java
@@ -82,11 +82,15 @@ public class NamespaceConfigurationService {
   protected void validateUpdate(NamespaceConfigurationApi existing,
       NamespaceConfigurationApi updated) {
     if (!Objects.equals(existing.getId(), updated.getId()) ||
-        !Objects.equals(existing.getAuth().getNamespace(), updated.getAuth().getNamespace())) {
+        !Objects.equals(existing.getAuth().getNamespace(), updated.getAuth().getNamespace()) ||
+        !Objects.equals(existing.getTaskQuotasConfiguration().getDetectionTaskQuota(),
+            updated.getTaskQuotasConfiguration().getDetectionTaskQuota()) ||
+        !Objects.equals(existing.getTaskQuotasConfiguration().getNotificationTaskQuota(),
+            updated.getTaskQuotasConfiguration().getNotificationTaskQuota())) {
       throw badRequest(
           ThirdEyeStatus.ERR_NAMESPACE_CONFIGURATION_VALIDATION_FAILED,
           existing.namespace(),
-          "Updating Id or auth is not allowed for Namespace Configuration");
+          "Updating Id, auth, or taskQuotasConfiguration is not allowed for Namespace Configuration");
     }
   }
 }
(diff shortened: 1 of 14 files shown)


thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/TaskQuotasConfigurationApi.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT * WARRANTIES OF ANY KIND,
+ * either express or implied.
+ * See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package ai.startree.thirdeye.spi.api;
+
+public class TaskQuotasConfigurationApi {
```
cyrilou242, line 16: let's have a first level named `QuotasConfiguration` where we will put all quotas.

anshul98ks123: makes sense. on it!

thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/api/TaskQuotasConfigurationApi.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private Long DetectionTaskQuota;
```
cyrilou242, line 18: let's have a more explicit name. quotas could be monthly rate, minutely rate, absolute number since account creation, per namespace, per user, etc...

also it's non standard to start a field name with an UPPER character. let's use lower character.

cyrilou242: no strong opinion on the name, but I'd go with something like

```
maximumDetectionTasksPerMonth 
```

anshul98ks123: ack

thirdeye-server/src/main/java/ai/startree/thirdeye/service/NamespaceConfigurationService.java
```diff
@@ -82,11 +82,15 @@ protected NamespaceConfigurationApi toApi(final NamespaceConfigurationDTO dto) {
   protected void validateUpdate(NamespaceConfigurationApi existing,
       NamespaceConfigurationApi updated) {
     if (!Objects.equals(existing.getId(), updated.getId()) ||
-        !Objects.equals(existing.getAuth().getNamespace(), updated.getAuth().getNamespace())) {
+        !Objects.equals(existing.getAuth().getNamespace(), updated.getAuth().getNamespace()) ||
+        !Objects.equals(existing.getTaskQuotasConfiguration().getDetectionTaskQuota(),
+            updated.getTaskQuotasConfiguration().getDetectionTaskQuota()) ||
```
cyrilou242, line 87: this will throw a NPE if updated.getTaskQuotasConfiguration() returns null. 
Let's override equals properly in the Quota DTO, (in intelliJ: code --> generate --> equals and hashCode)
then let's just use

```
Objects.equals(existing.getTaskQuotasConfiguration(), updated.getTaskQuotasConfiguration())
```

this will manage the case where .getTaskQuotasConfiguration() returns null and check all the fields, so it'll be more future proof if we ever add fields to the QuotaDto. (as long as `equals` is maintained correctly in the DTO but it's standard practice)

anshul98ks123: nice!
do we need to do this for the underlying taskQuotasConfigurationApi as well?

anshul98ks123: but this equals will still fail if the object pointers are different
which will likely be the case since we are comparing two different objects

we need a way to just compare the values and not the pointers of objects or underlying fields

anshul98ks123: Screenshot 2025-01-15 at 8 54 50 PM

anshul98ks123: > his will throw a NPE if updated.getTaskQuotasConfiguration() returns null

addressed all other feedback
will handle this one as a separate commit

do let me know if there are any other concerns

cyrilou242: > but this equals will still fail if the object pointers are different
which will likely be the case since we are comparing two different objects

> we need a way to just compare the values and not the pointers of objects or underlying fields

please read
https://www.baeldung.com/java-equals-hashcode-contracts

and Effective Java items 10, 11, 12, 13

see an example of overriding here: 
https://github.com/startreedata/thirdeye/blob/c3fa2926353aee7c495e9b21db8a0eb15f5193ef/thirdeye-spi/src/main/java/ai/startree/thirdeye/spi/datalayer/dto/AnomalyDTO.java#L336
but no need to write manually you can generate it, see message above

PR #1759, opened 2025-01-17
[spi] add TaskSubType

cyrilou242:
See https://startree.atlassian.net/browse/TE-2628

The frontend needs more information about the tasks to share information like: 
- an onboarding task is currently running
- a reset task following an upgrade is currently running
- etc...

Add TaskSubType to provide this information to the backend.


anshul98ks123 (approved):
:shipit:

PR #1775, opened 2025-01-24
fix(ui): TE-2635 fix chart load issue on date change in alert details

nalin-patidar:
#### Issue(s)

https://startree.atlassian.net/browse/TE-2635

#### Description

Chart data was not loading when user was updating the date range in alert details page.

#### Checklist

#### Screenshots

#### How to test


cyrilou242 (changes requested):
can we have at least 1 flow in the e2e tests for time picker usage ? 
if possible all flows where a time picker is visible.

PR #1800, opened 2025-02-17
[persistence] Optimize anomalies list API with combined query

anshul98ks123:
#### Issue(s)

#### Description

#### Checklist

#### Screenshots

#### How to test

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/AnomalyManagerImpl.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/AnomalyManagerImpl.java
@@ -74,9 +74,12 @@ public class AnomalyManagerImpl extends AbstractManagerImpl<AnomalyDTO>
   private static final ExecutorService EXECUTOR_SERVICE = Executors.newFixedThreadPool(10,
       new ThreadFactoryBuilder().setNameFormat("anomaly-manager-%d").build());
 
+  protected final GenericPojoDao genericPojoDao;
+
   @Inject
   public AnomalyManagerImpl(final GenericPojoDao genericPojoDao) {
     super(AnomalyDTO.class, genericPojoDao);
+    this.genericPojoDao = genericPojoDao;
   }
 
   @Override
@@ -195,7 +198,8 @@ public class AnomalyManagerImpl extends AbstractManagerImpl<AnomalyDTO>
 
   @Override
   public List<AnomalyDTO> filter(final DaoFilter daoFilter) {
-    final List<AnomalyDTO> anomalies = super.filter(daoFilter);
+    final List<AnomalyDTO> anomalies = genericPojoDao.getV2(
+        daoFilter.setBeanClass(AnomalyDTO.class));
     // FIXME CYRIL this filter is only decorating with feedback - while some others decorate with feedback and children
     return decorateWithFeedback(anomalies);
   }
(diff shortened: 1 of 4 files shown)


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/AnomalyManagerImpl.java
```diff
@@ -195,7 +198,8 @@ public List<AnomalyDTO> findAll() {
 
   @Override
   public List<AnomalyDTO> filter(final DaoFilter daoFilter) {
-    final List<AnomalyDTO> anomalies = super.filter(daoFilter);
+    final List<AnomalyDTO> anomalies = genericPojoDao.getV2(
```
cyrilou242, line 201: can we override `filter` in this class instead of making direct use of GenericPojoDao genericPojoDao ? 

then we will move the override to AbstractManagerImpl once we have confidence it works well

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
```diff
@@ -399,4 +400,55 @@ public <E extends AbstractDTO> int deleteByPredicate(final Predicate predicate,
...
+        daoFilter.getBeanClass());
+    validate(daoFilter);
+
+    final String matchingIdsQuery = databaseOrm.generateMatchingIdsQuery(
+        daoFilter.getPredicate(),
+        daoFilter.getLimit(),
+        daoFilter.getOffset(),
+        indexClass);
+    final String parameterizedSQL = String.format("""
+            JOIN (
```
cyrilou242, line 425: is this a hack to re-use the existing code ? 

is the perf different from ? 
```
generic_json_entity.id IN (<matchingIdsQuery>)
```
or is query plan the same

cyrilou242 (changes requested):
nice work!
put a few questions

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
anshul98ks123: IN queries do not support Limit and Offset in the inner Query
Hence had to move to JOIN query

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/AnomalyManagerImpl.java
anshul98ks123: We are already overriding filter in this class

AbstractManagerImpl has
```
@Override
  public List<E> filter(final DaoFilter daoFilter) {
    return genericPojoDao.get(daoFilter.setBeanClass(dtoClass));
  }
```

and we are overriding this here to use getV2 instead

do you mean that we should introduce a new method in AbstractManagerImpl and then use this here?

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
cyrilou242: > IN queries do not support Limit and Offset in the inner Query

TIL, thanks

cyrilou242 (approved):
:shipit:

thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/bao/AnomalyManagerImpl.java
cyrilou242: I missed that.
anyway code after your change looks good thanks

PR #1804, opened 2025-02-19
[persistence] Optimize result parsing in GenericPojoDao::getV2

anshul98ks123:
#### Issue(s)

[Investigate and fix anomalies list API slow response](https://startree.atlassian.net/browse/TE-2650)

#### Description

Optimizes parsing logic of query result in getV2 method of GenericPojoDao

#### Testing

- [X] Existing UTs & Integration Tests
- [X] Manual test

### DIFF
--- a/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
+++ b/thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
@@ -440,10 +440,11 @@ public class GenericPojoDao {
               Collections.emptyMap(),
               GenericJsonEntity.class,
               connection));
-      final List<E> results = new ArrayList<>();
+      final List<E> results = new ArrayList<>(entities.size());
       if (CollectionUtils.isNotEmpty(entities)) {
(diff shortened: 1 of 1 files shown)


thirdeye-persistence/src/main/java/ai/startree/thirdeye/datalayer/dao/GenericPojoDao.java
```diff
@@ -440,10 +440,11 @@ public <E extends AbstractDTO> List<E> getV2(final DaoFilter daoFilter) {
               Collections.emptyMap(),
               GenericJsonEntity.class,
               connection));
-      final List<E> results = new ArrayList<>();
+      final List<E> results = new ArrayList<>(entities.size());
       if (CollectionUtils.isNotEmpty(entities)) {
+        Class<E> beanClass = (Class<E>) daoFilter.getBeanClass();
```
cyrilou242, line 445: `final` here is actually important, it may help the compiler a lot

btw were you able to measure the cost of the casting ? in this case JIT should be able to figure this out
anyway code looks good.

let's have a working session about benchmarking and profiling if you're interested

cyrilou242 (approved):

